PR #7, opened 2025-08-03
Merge branch `develop` into branch `main`.

Davi0kProgramsThings:


src/protocols/index.js
chetasr, line 1: can we move the protocols into the generic plugin-style classes as described in the proposal? we can have the following repos created for the structure:

```
- wdk-protocol-base // contains base class for protocols + types
- wdk-protocol-lending // contains lending-protocol.js + types
- wdk-protocol-bridge // contains bridge-protocol.js + types
- wdk-protocol-swap // contains swap-protocol.js + types
```

src/wallet-account.js
chetasr, line 1: i also noticed that this class isn't really inherited anywhere, do we need it?

src/wallet-account-read-only.js
```diff
@@ -0,0 +1,190 @@
...
+  async getTransactionReceipt (hash) {
+    throw new NotImplementedError('getTransactionReceipt(hash)')
+  }
+}
+
+/** 
+ * @abstract
+ * @implements {IWalletAccountReadOnly}
+ */
+export default class WalletAccountReadOnly {
```
vigan-abd, line 108: question why this class is simply not extending IWalletAccountReadOnly and overriding necessary methods?

src/wallet-account.js
```diff
@@ -0,0 +1,117 @@
...
+/** @typedef {import('./wallet-account-read-only.js').TransferResult} TransferResult */
+
+/**
+ * @typedef {Object} KeyPair
+ * @property {Uint8Array} publicKey - The public key.
+ * @property {Uint8Array} privateKey - The private key.
+ */
+
+/** @interface */
+export class IWalletAccount extends IWalletAccountReadOnly {
```
vigan-abd, line 33: also we have WalletAccountReadOnly but not WalletAccount, so maybe we can simplify above to have just IWalletAccount? Or should we extend WalletAccount as well? not sure but inheritance here doesn't feel right, I would suggest to have IWalletAccount and IWalletAccountReadOnly only, let me know your thoughts

tests/wallet-account-read-only.test.js
```diff
@@ -0,0 +1,48 @@
+import { describe, expect, test } from '@jest/globals'
```
vigan-abd, line 1: not for now, but in future when we have time we should rewrite tests in brittle, we should try to promote hp team stack whenever possible

vigan-abd (commented):
some comments from my side as well regarding inheritance

package.json
```diff
@@ -0,0 +1,35 @@
...
+  "types": "./types",
+  "scripts": {
+    "build:types": "rimraf *.d.ts **/*.d.ts; tsc",
+    "lint": "standard",
+    "lint:fix": "standard --fix",
+    "test": "cross-env NODE_OPTIONS=--experimental-vm-modules jest",
+    "test:coverage": "cross-env NODE_OPTIONS=--experimental-vm-modules jest --coverage"
+  },
+  "dependencies": {
+    "bip39": "^3.1.0"
```
vigan-abd: should we have fixed version here?

src/wallet-account-read-only.js
```diff
@@ -0,0 +1,190 @@
...
+   * @returns {Promise<string>} The account's address.
+   */
+  async getAddress () {
+    throw new NotImplementedError('getAddress()')
+  }
+
+  /**
+   * Returns the account's native token balance.
+   *
+   * @returns {Promise<number>} The native token balance.
```
vigan-abd: question regarding this, I see we return number, is this decimal or unit currency, if it's unit we have to ensure we have it everywhere as string to have high precision, bigint would be ideal but introduces issues with json serialization

src/protocols/index.js
Davi0kProgramsThings: This was the initial idea, but me and dario realized that it wasn't really worth splitting each protocol base class into its own module as there isn't much content to justify it. It's way better to limit the number of repositories in such case. We still split the wallet and protocols definitions in two different roots: @wdk/wallet and @wdk/wallet/protocols (which we think is a good compromise that will benefit both us and our end users).

There isn't a common interface between the different protocol base classes, so we can't really define a base class that all of them should extend. We can still have a base class for each kind of protocols e.g. swap, bridge, lending, ..., as all swap protocos will have a swap and quoteSwap methods, all bridge protocols will have a bridge and quoteBridge methods, and so on...

src/wallet-account.js
Davi0kProgramsThings: The IWalletAccount class is the javascript equivalent of an interface, as it has no constructors and all its methods are abstract (which in javascript means that they will always throw an error indicating that the method must be implemented).

For such classes, the following implementations are both possible and equivalent (as when we implement the IWalletAccount interface, we always override all of its methods so actual inheritance isn't really necessary):
```javascript
// src/wallet-account-*.js
// Option 1:
export default class WalletAccount* extends IWalletAccount {

// Option 2:
/** @implements {IWalletAccount} */
export default class WalletAccount* {
```

So, why did we choose option 2. over option 1.? 

The first reason is that option 2. allows tsc to correctly generate the type definitions of the wallet modules, as IWalletAccount is defined as a typescript interface in the types of @wdk/wallet. So if we went with option 1., tsc would generate the following invalid .d.ts files for the wallet account classes:
```typescript
// types/wallet-account-*.d.ts
// Error: cannot extend an interface 'IWalletAccount'. Did you mean 'implements'? (2689)
export default class WalletAccount* extends IWalletAccount {
```

The second reason is that javacript doesn't support multiple inheritance, so we wouldn't be able to proceed with option 1. as all wallet account classes already extend their read-only implementations. Instead, option 2. allows us to both extend the read-only wallet account classes and implement the IWalletAccount interface (that includes definitions for the other non read-only methods):
```typescript
// types/wallet-account-*.d.ts
export default class WalletAccount* extends WalletAccountReadOnly* implements IWalletAccount {
```

The only bonus of option 1. is that if the developer forgets to implement one of the methods and a user tries to call it, he will get a not implemented error instead of "TypeError: account.* is not a function". This is insignificant in our wallet modules though, as we are always overriding all the methods (including the ones that the target blockchain doesn't support, e.g. transfer on @wdk/wallet-btc). 

Note that text editors and ides treat options 1. and 2. equivalently, so there are really no differences between the two.

src/wallet-account-read-only.js
Davi0kProgramsThings: Take a look at the following comment for a complete explanation: https://github.com/tetherto/wdk-wallet/pull/7#discussion_r2270886649.

src/wallet-account.js
Davi0kProgramsThings: We don't have a WalletAccount abstract class* that all wallet account classes extend just because they don't share any common code. There isn't a common constructor and there aren't any common method implementations between them, they only share the methods and their specifications. In such cases, an abstract class would look exactly like an interface (as it would have no constructor and no concrete methods). That's why we only need the IWalletAccount interface. 

Instead, all read-only wallet account classes have a common constructor and implementation for the 'getAddress' method that we can easily group in an abstract class. But why do we need both an IWalletAccountReadOnly interface and a WalletAccountReadOnly abstract class?

Basically, we need the IWalletAccountReadOnly interface so that IWalletAccount can extend it and add specifications for the missing non read-only methods. We could have just made the IWalletAccount interface extend the WalletAccountReadOnly abstract class but that would have been wrong since interfaces shouldn't include any code. It would have also led tsc to generate broken type definitions, as typescript doesn't allow interfaces to extend abstract classes. 

Note that the WalletAccountReadOnly abstract class implements the IWalletAccountReadOnly interface. This is actually a really common design pattern, as it lets users choose whether to implement the interface from scratch or by extending the already existing abstract class (which provides some ready-to-use boilerplate).

\* Javascript doesn't support real interfaces and abstract classes as typescript does, however it's still possible to emulate them. In fact, all classes that include methods that always throw a not implemented error are conceptually abstract classes (or at least, they behave exactly like one). In @wdk/wallet, WalletManager and WalletAccountReadOnly are examples of javascript abstract classes. By marking them with the jsdoc's @abstract tag, tsc will also generate the proper typescript abstract classes equivalent in the .d.ts files:
```typescript
// src/wallet-manager.js
/** @abstract */
export default class WalletManager {
  [...]

  /**
   * Returns the wallet account at a specific index (see [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)).
   *
   * @abstract
   * @param {number} [index] - The index of the account to get (default: 0).
   * @returns {Promise<IWalletAccount>} The account.
   */
  async getAccount (index = 0) {
    throw new NotImplementedError('getAccount(index)')
  }
}

// types/wallet-manager.d.ts
export default abstract class WalletManager {
  [...]

  /**
   * Returns the wallet account at a specific index (see [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)).
   *
   * @param {number} [index] - The index of the account to get (default: 0).
   * @returns {Promise<IWalletAccount>} The account.
   */
   abstract getAccount(index?: number): Promise<IWalletAccount>;
}
```

tests/wallet-account-read-only.test.js
Davi0kProgramsThings: 👍🏻

package.json
Davi0kProgramsThings: My bad, i will fix it soon.

src/wallet-account-read-only.js
Davi0kProgramsThings: We already have in plan to change all numeric return values from number type to bigint. We will also extend the type of all numeric arguments to number | bigint in order to ensure no precision loss while allowing users to still use numbers without first needing to convert them to bigint.

The json serialization issue with bigint values won't really be a problem here. It never happens in the rumble wallet application and generally it is really easy to provide an encoder to serialize big integers into a string format.

src/wallet-account.js
vigan-abd: thanks for detailed clarifications behind this, makes more sense to me as well now

src/wallet-account-read-only.js
vigan-abd: ty, makes sense

src/wallet-account.js
vigan-abd: ty, makes sense

src/wallet-account-read-only.js
vigan-abd: ok then agree with using bigint on responses and number | bigint in inputs as long as we ensure wrapping stuff on bigint

vigan-abd (dismissed):

vigan-abd:
would be good to address this soon after merge https://github.com/tetherto/wdk-wallet/pull/7#discussion_r2272903631

vigan-abd (approved):

chetasr (approved):

subashsn (approved):

chetasr (approved):

PR #23, opened 2025-11-15
Poc signer

claudiovb:
# Description

Extend WalletManager to support a signer object in addition to the existing seed.

The constructor now accepts either a BIP-39 seed (string mnemonic or raw `Uint8Array`) or an `ISigner` instance, determined at runtime via type detection. This preserves full backwards compatibility with the seed-based workflow while enabling the new signer-based architecture.

### Changes

- **WalletManager constructor**: accepts `string | Uint8Array | ISigner` as the first argument
- **`get seed()`**: re-added for consumers using the seed-based path
- **`ISigner` interface**: new base class for chain-agnostic signers (`derive`, `getAddress`, `dispose`)
- **`createSigner` / `getSigner`**: methods for managing multiple named signers
- **`getAccount` / `getAccountByPath`**: now accept an optional `signerName` parameter

## Motivation and Context

We need to enable the WDK modules to support multiple different signers (hardware devices, remote signers, etc.) while maintaining backwards compatibility with existing seed-based wallet implementations.

## Type of change

---
  - https://app.asana.com/0/0/1213909416546895


src/wallet-manager.js
```diff
@@ -141,6 +190,13 @@ export default class WalletManager {
       }
     }
 
+    for (const signer of Object.values(this._signers)) {
+      if (signer.keyPair?.privateKey) {
+        signer.dispose()
+      }
```
AlonzoRicardo: why not unconditionally call dispose() on every signer?

claudiovb: Good point,the privateKey check is redundant here. Each dispose() implementation is expected to be idempotent (safe to call multiple times), so the caller shouldn't need to guard against that. Removed the checks.

package.json
```diff
@@ -1,8 +1,12 @@
 {
   "name": "@tetherto/wdk-wallet",
-  "version": "1.0.0-beta.7",
+  "version": "2.0.0-beta.0",
```
AlonzoRicardo: wasn't this going to stay a beta bump? i remember the thread back in april where we agreed major bumps don't really fit a project that's still in beta, especially since this PR is backwards compatible now. should this be `1.0.0-beta.8` instead?

tests/wallet-manager.test.js
```diff
@@ -100,4 +107,13 @@ describe('WalletManager', () => {
         .toBe(false)
     })
   })
+
+  describe('getSigner', () => {
+    test('should return the default signer when only default signer exists', () => {
```
AlonzoRicardo: isn't this the same test as the one on line 54? both create a signer and check `getSigner(default)` returns it. maybe drop one or make this one test a non-default signer?

src/wallet-manager.js
```diff
@@ -91,22 +117,44 @@ export default class WalletManager {
...
+    throw new NotImplementedError('createSigner(signerName, signer)')
+  }
+
+  /**
+   * Returns a signer.
+   *
+   * @param {string} signerName - The signer name.
+   * @returns {ISigner} The signer.
+   */
+  getSigner (signerName) {
```
AlonzoRicardo, line 158: should we throw here if the signer doesn't exist? right now consumers get `undefined` with no hint, and `getAccountByPath` already throws so the behavior is inconsistent

package.json
claudiovb: True it slipped to me the reversion here , tks. I reverted on the merge already. Also I left the same as the current version because I think we update only on release PRs automatically now

AlonzoRicardo (dismissed):

src/signer.js
Davi0kProgramsThings, line 1: Rename this file to src/signer.js.

src/signer.js
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Chain-specific signers (EVM, BTC, Tron, Solana, etc.) can extend this
+ * contract with additional capabilities (e.g., signTransaction, signTypedData,
+ * signPsbt), but those are intentionally not included here to keep the base
+ * interface chain-agnostic.
+ *
+ * Common optional fields/methods implementers may provide:
+ *  - index: number | undefined
+ *  - path: string | undefined
+ *  - address: string | undefined
```
Davi0kProgramsThings: There's no real need to include this in the description of the class.

src/signer.js
```diff
@@ -0,0 +1,71 @@
...
+ *  - dispose: clear any secret material from memory
+ *
+ * Chain-specific signers (EVM, BTC, Tron, Solana, etc.) can extend this
+ * contract with additional capabilities (e.g., signTransaction, signTypedData,
+ * signPsbt), but those are intentionally not included here to keep the base
+ * interface chain-agnostic.
+ *
+ * Common optional fields/methods implementers may provide:
+ *  - index: number | undefined
+ *  - path: string | undefined
```
Davi0kProgramsThings: What's the reason we are not including these properties in the common interface? All signers should be able to provide at least the path.

src/signer.js
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @interface
+ */
+export class ISigner {
+  /**
+   * Derive a child signer using a relative path.
+   *
+   * Example formats (chain-dependent):
+   *  - BIP-32/BIP-44 relative paths like "0'/0/0"
+   *  - Implementation-specific segmenting, as applicable
```
Davi0kProgramsThings: I guess this can just be:
```suggestion
   * Derive a child signer using a relative path (e.g., "0'/0/0").
```

src/signer.js
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * @param {string} relPath - The relative derivation path.
+   * @param {WalletConfig} [cfg] - Optional chain-specific configuration.
```
Davi0kProgramsThings: It's probably better to use unknown here instead of 'WalletConfig', since the latter includes only the 'transferMaxFee' property which doesn't really make much sense in this context.
```suggestion
   * @param {unknown} [config] - Optional chain-specific configuration.
```

src/signer.js
```diff
@@ -0,0 +1,71 @@
...
+   * @returns {ISigner} The derived signer.
+   */
+  derive (relPath, cfg) {
```
Davi0kProgramsThings: Are we sure we don't want to make this method asynchronous? There might be some signers that requires async operations to be run in order to derive a path, and by making the method synchronous we wouldn't support them.

src/signer.js
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * @returns {Promise<string>} The address.
+   */
+  async getAddress () {
+    throw new NotImplementedError('getAddress()')
+  }
+
+  /**
+   * Disposes the signer and clears any secret material from memory.
+   * Implementations should be idempotent (safe to call more than once).
```
Davi0kProgramsThings: Where does this come from? After calling 'dispose', the user should always expect undefined behavior when interacting with the object. It's not really necessary for the 'dispose' method to be idempotent so it's probably better to get rid of this statement.

src/wallet-manager.js
```diff
@@ -35,29 +36,46 @@ export default class WalletManager {
   /**
    * Creates a new wallet manager.
    *
-   * @param {string | Uint8Array} seed - The wallet's [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) seed phrase.
+   * Accepts either a BIP-39 seed (string mnemonic or raw Uint8Array) for
+   * backwards compatibility, or an {@link ISigner} instance for the new
+   * signer-based workflow.
```
Davi0kProgramsThings: No need to state that BIP-39 seeds are for backwards compatibility:
```suggestion
   * Accepts either a BIP-39 seed (string mnemonic or raw Uint8Array), or 
   * an {@link ISigner} instance.
```

src/wallet-manager.js
```diff
@@ -67,6 +85,14 @@ export default class WalletManager {
...
      */
     this._accounts = {}
+
+    /**
+     * The wallet configuration.
+     *
+     * @protected
+     * @type {WalletConfig}
+     */
+    this._config = config
```
Davi0kProgramsThings, line 98: Avoid unnecessary or out-of-scope changes.

src/wallet-manager.js
```diff
@@ -35,29 +36,46 @@ export default class WalletManager {
...
+      this._seed = null
+
+      /**
+       * A map between signer names and signers.
+       *
+       * @protected
+       * @type {{ [name: string]: ISigner }}
+       */
+      this._signers = { default: seedOrSigner }
     }
```
Davi0kProgramsThings: You can refactor this code to:
```javascript
  constructor (seedOrSigner, config = { }) {
    if (typeof seedOrSigner === 'string') {
      if (!WalletManager.isValidSeedPhrase(seedOrSigner)) {
        throw new Error('The seed phrase is invalid.')
      }

      seedOrSigner = bip39.mnemonicToSeedSync(seedOrSigner)
    }

    /** @private */
    this._seed = seedOrSigner instanceof Uint8Array ? seedOrSigner : undefined

    /**
     * A map between signer names and signers.
     *
     * @protected
     * @type {Record<string, ISigner>}
     */
    this._signers = seedOrSigner instanceof Uint8Array ? { } : { default: seedOrSigner }

    [...]
```

src/wallet-manager.js
```diff
@@ -35,29 +36,46 @@ export default class WalletManager {
...
 
-      seed = bip39.mnemonicToSeedSync(seed)
+      /** @private */
+      this._seed = seedOrSigner
+
+      /**
+       * A map between signer names and signers.
+       *
+       * @protected
+       * @type {{ [name: string]: ISigner }}
```
Davi0kProgramsThings: Use 'Record<string, ISigner>':
```suggestion
       * @type {Record<string, ISigner>}
```

src/wallet-manager.js
```diff
@@ -91,34 +117,62 @@ export default class WalletManager {
   }
 
   /**
-   * The seed phrase of the wallet.
+   * The seed of the wallet.
    *
-   * @type {Uint8Array}
+   * @type {Uint8Array | null}
```
Davi0kProgramsThings: It's probably better to use undefined instead of null to indicate that the seed doesn't exist:
```suggestion
   * @type {Uint8Array | undefined}
```

src/wallet-manager.js
```diff
@@ -91,34 +117,62 @@ export default class WalletManager {
...
+  /**
+   * Creates a new signer.
+   *
+   * @abstract
+   * @param {string} signerName - The signer name.
+   * @param {ISigner} signer - The signer.
+   */
+  createSigner (signerName, signer) {
+    throw new NotImplementedError('createSigner(signerName, signer)')
+  }
```
Davi0kProgramsThings, line 146: We should be able to easily implement this method instead of making it abstract. Also, what's the exact rationale behind this feature? Users could just create separate instances that use different signers instead of doing everything with a single object.

src/wallet-manager.js
```diff
@@ -91,34 +117,62 @@ export default class WalletManager {
...
+
+  /**
+   * Returns a signer registered under the given name.
+   *
+   * @param {string} signerName - The signer name.
+   * @returns {ISigner} The signer.
+   * @throws {Error} If no signer is registered under `signerName`.
+   */
+  getSigner (signerName) {
```
Davi0kProgramsThings, line 158: The 'signerName' argument should be optional. If the callee doesn't provide it, then the method should:
1. if the user provided a default signer at construction, return the default signer.
2. if the user didn't provide a default signer at construction, throw an error.

src/wallet-manager.js
```diff
@@ -35,29 +36,46 @@ export default class WalletManager {
...
+      /** @private */
+      this._seed = null
+
+      /**
+       * A map between signer names and signers.
+       *
+       * @protected
+       * @type {{ [name: string]: ISigner }}
+       */
+      this._signers = { default: seedOrSigner }
```
Davi0kProgramsThings: In order to keep this an implementation detail, and avoid users overwriting it by mistake use something like _default as the key for the default signer or even better use a separate _defaultSigner property.

src/wallet-manager.js
```diff
@@ -91,34 +117,62 @@ export default class WalletManager {
...
+    }
+    return signer
+  }
+
   /**
    * Returns the wallet account at a specific index (see [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)).
    *
    * @abstract
    * @param {number} [index] - The index of the account to get (default: 0).
+   * @param {string} [signerName='default'] - The name of the signer to use.
```
Davi0kProgramsThings: Instead of using a positional argument, add an 'options' parameter that includes a 'signerName: string' field. Also, the default value should be undefined for the same reason of the 'getSigner' method:
```suggestion
   * @param {Object} [options] - The signer options.
   * @param {string} [options.signerName] - The name of the signer to use.
```

src/wallet-manager.js
```diff
@@ -132,15 +186,18 @@ export default class WalletManager {
...
   /**
-   * Disposes all the wallet accounts, erasing their private keys from the memory.
+   * Disposes all wallet accounts and signers, clearing secret material from memory.
    */
   dispose () {
     for (const account of Object.values(this._accounts)) {
-      if (account.keyPair.privateKey) {
-        account.dispose()
-      }
+      account.dispose()
```
Davi0kProgramsThings: We need the if statement to ensure that no errors are thrown in the case one of the accounts has already been disposed separately.

src/wallet-manager.js
```diff
@@ -91,34 +117,62 @@ export default class WalletManager {
...
   }
 
+  /**
+   * Creates a new signer.
+   *
+   * @abstract
+   * @param {string} signerName - The signer name.
+   * @param {ISigner} signer - The signer.
+   */
+  createSigner (signerName, signer) {
```
Davi0kProgramsThings: Rename this method to:
```suggestion
  addSigner (signerName, signer) {
```

src/wallet-manager.js
```diff
@@ -91,34 +117,62 @@ export default class WalletManager {
...
+    return signer
+  }
+
   /**
    * Returns the wallet account at a specific index (see [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)).
    *
    * @abstract
    * @param {number} [index] - The index of the account to get (default: 0).
+   * @param {string} [signerName='default'] - The name of the signer to use.
    * @returns {Promise<IWalletAccount>} The account.
```
Davi0kProgramsThings, line 190: Missing @throws:
```suggestion
   * @returns {Promise<IWalletAccount>} The account.
   * @throws {Error} If a signer name is given, but no signer exists with the given name.
```

tests/wallet-manager.test.js
```diff
@@ -1,9 +1,20 @@
...
 
-import { describe, expect, test } from '@jest/globals'
+import { describe, expect, jest, test } from '@jest/globals'
 
 import WalletManager from '../index.js'
 
+/**
+ * @param {string} [name]
+ * @returns {{ name: string, dispose: ReturnType<typeof jest.fn> }}
+ */
```
Davi0kProgramsThings: No need to type or document tests.

tests/wallet-manager.test.js
```diff
@@ -15,13 +26,13 @@ class DummyWalletManager extends WalletManager {
...
-      fast: 0
+      normal: 0n,
+      fast: 0n
     }
   }
 
   dispose () {
-
+    super.dispose()
   }
```
Davi0kProgramsThings: You can just remove the override of the 'dispose' method.

tests/wallet-manager.test.js
```diff
@@ -100,4 +124,29 @@ describe('WalletManager', () => {
...
+      const wallet = new DummyWalletManager(signer)
+
+      expect(wallet.getSigner('default')).toBe(signer)
+
+      wallet.dispose()
+
+      expect(signer.dispose).toHaveBeenCalledTimes(1)
+      expect(() => wallet.getSigner('default'))
+        .toThrow('No signer registered with name "default".')
+    })
```
Davi0kProgramsThings, line 185: Remove this test case.

src/signer.js
Davi0kProgramsThings: Rename this argument to options:
```suggestion
   * @param {unknown} [options] - Optional chain-specific configuration.
```

tests/wallet-manager.test.js
```diff
@@ -1,9 +1,20 @@
...
-import { describe, expect, test } from '@jest/globals'
+import { describe, expect, jest, test } from '@jest/globals'
 
 import WalletManager from '../index.js'
 
+/**
+ * @param {string} [name]
+ * @returns {{ name: string, dispose: ReturnType<typeof jest.fn> }}
+ */
+function createDummySigner (name = 'dummy') {
```
Davi0kProgramsThings: Instead of using a 'createDummySigner' function, use:
```javascript
class DummySigner {
  derive(relPath, options) {
    return this
  }

  async getAddress() {
    return 'dummy-address'
  }

  dispose () {
  
  }
}
```

types/src/wallet-manager.d.ts
```diff
@@ -1,3 +1,14 @@
...
+/** @typedef {import('./isigner.js').ISigner} ISigner */
+/**
+ * @typedef {Object} WalletConfig
+ * @property {number | bigint} [transferMaxFee] - The maximum fee amount for transfer operations.
+ */
+/**
+ * @typedef {Object} FeeRates
+ * @property {bigint} normal - The fee rate for transaction sent with normal priority.
+ * @property {bigint} fast - The fee rate for transaction sent with fast priority.
+ */
```
Davi0kProgramsThings: Remove these comments at the beginning of the file.

src/wallet-manager.js
```diff
@@ -35,29 +36,46 @@ export default class WalletManager {
...
+   * signer-based workflow.
+   *
+   * @param {string | Uint8Array | ISigner} seedOrSigner - A BIP-39 seed phrase, raw seed bytes, or a default signer.
    * @param {WalletConfig} [config] - The wallet configuration.
    */
-  constructor (seed, config = { }) {
-    if (typeof seed === 'string') {
-      if (!WalletManager.isValidSeedPhrase(seed)) {
-        throw new Error('The seed phrase is invalid.')
+  constructor (seedOrSigner, config = { }) {
```
Davi0kProgramsThings: Instead of making the type of the 'seedOrSigner' argument a union of three types, we should provide two constructor overloads: one for BIP-39 seeds, and one for signers. The implementation will remain the same, so it's just a matter of specification. It should be quite similar to the following:
https://github.com/tetherto/wdk-wallet-evm-7702-gasless/blob/main/src/wallet-account-evm-7702-gasless.js#L56

Davi0kProgramsThings (changes requested):

src/signer.js
claudiovb: The reason is because adding `path` to the common interface forces implementations that don't have one to lie. For example a signer built from a raw private key, imagine importing a key exported from MetaMask, a hardware backup, or any non-HD flow those do not have no derivation path. Forcing it to surface `path: ''` or `path: null` is worse than absence: a caller that reads it could end up signing under or deriving from a wrong path.

`path` and `index` are BIP-32 / BIP-44 concepts, not universal signer concepts. Keeping them as documented optional fields lets HD signers expose them honestly while non-HD signers stay correct.

src/wallet-manager.js
claudiovb: `_config` isn't dead, subclasses that inherit from `WalletManager` rely onit. Example in [wdk-wallet-btc](https://github.com/claudiovb/wdk-wallet-btc/blob/f10be26ccd9cf46e9be9b3bddb30c3a18892db07/src/wallet-manager-btc.js#L62). 
We need it stored there because when deriving new accounts or even with the new failoverProviders we need to pass again these configs when new objects are created...
The base class is the right place to own that storage so each subclass doesn't re-implement it. I'll can add short JSDoc making the intent explicit.

src/wallet-manager.js
claudiovb: The main value of `createSigner` is sharing one manager so values such as network, client, account map, config can be reused across multiple signers without standing up a parallel instance per signer. Use cases:

- Seed + hardware wallet plugged in mid-session
- Seed + an externally imported private key
- Two seeds (e.g., personal + corporate) under the same network setup
- Account-abstraction-style flows mixing an EOA signer with a smart-account
  signer

The alternative, one `WalletManager` per signer, works, but duplicates the network/client setup and forces a teardown/rebuild every time the user adds a key or pairs a device. Hardware wallets in particular are usually paired after construction, not at it.

You're right that it doesn't need to be abstract, will push a fix making it concrete tks.

src/wallet-manager.js
claudiovb: We actually went the other way earlier in this PR, @AlonzoRicardo  raised the same question [here](https://github.com/tetherto/wdk-wallet/pull/23#discussion_r3040005417) and we landed on dropping the guard: each `dispose()` implementation getting dispose to be idempotent, so the responsibility lives with the leaf (signer / account), not with the manager. Re-adding the `if` here pushes that responsibility back up and re-introduces the inconsistency Alonzo flagged.
We can add it back, no problems with me but just flagging this here so we do keep going back and forth on this point. 
I validated the idempotent behaviour on the btc and evm classes with the signers version at least but indeed if this would cause issues with some other class might be worth reverting it

src/signer.js
claudiovb: There's a comment down below in a subject related to this. Probably worth solving direction there before proceeding here

tests/wallet-manager.test.js
claudiovb: This test covers behavior introduced by this PR, `dispose()` now delegates to signers and clears the signer map, neither of which existed before. It felt worth pinning. No strong attachment though, happy to drop it if you'd still rather it goes. Also I fixed it per the previous comments already

types/src/wallet-manager.d.ts
claudiovb: Mostly a stylistic preference keeping these `@typedef` blocks at the top mirrors the JSDoc style we use throughout the source `.js` files and gives a quick index of what types the module pulls in without scrolling to the bottom. They overlap with the `export type` declarations at the end, so they're not strictly load-bearing, but I'd rather keep the types listed up-front the same way the source files do (also if we ever get back building the types automatically with tsc command they add the jsdocs comments on the ts file generated too...). Open to dropping if you feel strongly though.

src/signer.js
claudiovb: Went with config instead of options to align with the surrounding API: the WalletManager constructor and the derive method both already use config for chain-specific configuration, so this keeps the naming consistent across the signer/manager surface. Happy to revisit if you'd prefer the entire surface move to options instead.

src/signer.js
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+'use strict'
+
+import { NotImplementedError } from './errors.js'
+
+/**
+ * A minimal, cross-chain signer interface. Chain-specific signers can extend
+ * this contract with additional capabilities (e.g., signTransaction, signPsbt),
+ * which are intentionally kept out of the base to remain chain-agnostic.
```
Davi0kProgramsThings: No need to state that the chain-specific signers can extend the class with additional capabilities, or that those features are intentionally left out of the base class in order to keep it chain-agnostic. That's a bunch of useless information that could just confuse the user.
```suggestion
 * A minimal, cross-chain signer interface.
```

src/signer.js
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+'use strict'
+
+import { NotImplementedError } from './errors.js'
+
+/**
+ * A minimal, cross-chain signer interface. Chain-specific signers can extend
+ * this contract with additional capabilities (e.g., signTransaction, signPsbt),
```
Davi0kProgramsThings: Any reasons we shouldn't include the sign transaction method in the common interface?

src/signer.js
Davi0kProgramsThings: c.c. @jonathunne 

We don't really need signers that accept a single private key since users can just use the new static factory methods. Also, i'm not sure we want to support non-HD flows since the wallet development kit is by definition an implementation of the BIP 0032 specification. Actually, i can't really see any use case that we might want to encourage and that doesn't make use of a path.

src/wallet-manager.js
Davi0kProgramsThings: My comment was just about moving the definition of the _config property to a different position in the constructor. That's an unnecessary change that just bloats the git diff. You don't need to revert it, but keep in mind to avoid unnecessary and out-of-scope changes in the future.

src/wallet-manager.js
```diff
@@ -33,40 +34,68 @@ import { NotImplementedError } from './errors.js'
...
+    const isSeed = seedOrSigner instanceof Uint8Array
+
     /** @private */
-    this._seed = seed
+    this._seed = isSeed ? seedOrSigner : undefined
 
     /**
-     * The wallet configuration.
+     * The default signer provided at construction. Accessed via {@link getSigner}
+     * with no arguments.
```
Davi0kProgramsThings: No need to state how to access the property:
```suggestion
     * The default signer.
```

src/wallet-manager.js
```diff
@@ -91,34 +120,79 @@ export default class WalletManager {
...
 
+  /**
+   * Registers a signer under the given name.
+   *
+   * @param {string} signerName - The signer name.
+   * @param {ISigner} signer - The signer.
+   * @throws {Error} If `signerName` is empty.
+   */
+  addSigner (signerName, signer) {
+    if (!signerName) {
```
Davi0kProgramsThings: The 'signerName' argument is not optional, so you don't need to check for nullability. You could still check and throw a proper error in the case the signer name is an empty or a blank string. This isn't strictly necessary, though.

src/wallet-manager.js
Davi0kProgramsThings: c.c. @jonathunne 

Wallet managers never actually take care of establishing connections with external services and networks. In all of our wallet modules, this stuff almost always happen inside the single wallet accounts. This doesn't slow down anything since connections are usually stateless and don't require any actual setup or teardown. Hardware wallets are a different matter, but drivers should provide some sort of stateless access as well. Even in the case they don't, the set up and tear down might be quick operations or we could just fix the issue by implementing a proper singleton class. In my opinion, this solves an issue that doesn't really exist at the cost of introducing ambiguity and encouraging bad and messy code. Also:

- Seed + hardware wallet plugged in mid-session
Why should users plug hardware wallets in mid-session? Also, why should they use a seed if they will connect an hardware wallet? This seems like an edge case. They could also just define and use two different wallet managers with no trade-offs here.

- Seed + an externally imported private key
This will never happen since private key signers will not exist. Users will be able to initialize wallet accounts for single private keys by using the new static factory methods:
    ```javascript
    const account = await WalletAccount*.fromPrivateKey('<your private-key>', config)
    ```

- Account-abstraction-style flows mixing an EOA signer with a smart-account signer
I don't think we will ever implement this? EOA and abstract accounts are from two different wallet modules, not signers. If users want to derive and use EOAs they should use wdk-wallet-evm, otherwise they should rely on wdk-wallet-evm-erc-4337 (or wdk-wallet-evm-7702-gasless).

src/wallet-manager.js
```diff
@@ -91,34 +120,79 @@ export default class WalletManager {
...
+    throw new NotImplementedError('getAccount(index, options?)')
   }
 
   /**
    * Returns the wallet account at a specific [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) derivation path.
    *
    * @abstract
    * @param {string} path - The derivation path (e.g. "0'/0/0").
+   * @param {Object} [options] - Account options.
+   * @param {string} [options.signerName] - The signer name. Omit to use the default signer.
```
Davi0kProgramsThings, line 218: This might be confusing since users can omit the field even without a default signer and the method would still run without throwing any error, since it would use the seed and path given at construction. For this reason, i would just get rid of that part:
```suggestion
   * @param {string} [options.signerName] - The signer name.
```

src/wallet-manager.js
Davi0kProgramsThings: This can definitely cause issues with other classes, so we should indeed revert it. Also, this change would require to update the dispose logic of all official and third-party wallet modules for no reason. The specification of the 'IWalletAccount' interface states that after disposal the private key is set to null, so the guard is perfectly fine. Also, by requiring the 'dispose' method to be idempotent you force additional logic that just makes implementations of the method more complex. E.g., classes that close a network connection in the 'dispose' method should add an extra guard to ensure that the connection is actually open.

src/signer.js
Davi0kProgramsThings: So, if i’ve got this right you will use this argument as the configuration object of the new signer that the method will derive. Correct? In such case, yes. It makes sense to use 'config'.

types/src/wallet-manager.d.ts
Davi0kProgramsThings: Honestly tsc behaves pretty weirdly with these comments. Sometimes it generates them, sometimes it doesn't. I like the idea of keeping them in order to provide some sort of index of the types that the module exports, but i would still get rid of them for the sake of consistency with the other modules whose type definitions don't include them.

Davi0kProgramsThings (changes requested):

src/signer.js
claudiovb: Yes , perfect - tks.

src/signer.js
claudiovb: Given the status of the comment below will drop this text from here as well

types/src/wallet-manager.d.ts
claudiovb: ok will drop them

src/wallet-manager.js
claudiovb: The if as is checks for this empty or blank string, even if not necessary I think its a nice plus here. Will leave it there as it is for now, if you think would be much better without the check I can drop it too.

src/wallet-manager.js
claudiovb: Added the if back this makes sense, tks

src/signer.js
claudiovb: Mostly because the sign-transaction operation looks completely different per chain -> shape, name, and even arity diverge:

**Shape diverges:**

EVM: typed object compatible with ethers.Transaction.from: [signTransaction(unsignedTx)](https://github.com/claudiovb/wdk-wallet-evm/blob/e615962e1c081357f45dc58167a94c29cf2d2ae7/src/signers/seed-signer-evm.js#L270)
Bitcoin: PSBT (Partially Signed Bitcoin Transaction), is a binary, multi-input, completely different function: [signPsbt(psbt)](https://github.com/claudiovb/wdk-wallet-btc/blob/f10be26ccd9cf46e9be9b3bddb30c3a18892db07/src/signers/seed-signer-btc.js#L386)
TON: cells / BOC
Solana: VersionedTransaction with instruction arrays
Names diverge too, and this is the bigger issue. Bitcoin's convention for "sign but don't broadcast" is signPsbt, not signTransaction. Putting signTransaction on the common interface is forcing every chain implementer to adopt Ethereum's naming convention regardless of what their own ecosystem calls the operation. That's a leaky abstraction.

Even the amount of sign functions required diverges or the arguments those should need: EVM's signTypedData(domain, types, message) takes three positional args; signPsbt(psbt) takes one; some chains might want two (e.g., tx + signing context). Blockchain implementations are a broad landscape, and locking down a signature shape on the base forces every implementer to either lie about it or wrap/unwrap on every call.

That's why the idea is that we split into chain-specific sub-interfaces that extend ISigner: [ISignerBtc](https://github.com/claudiovb/wdk-wallet-btc/blob/f10be26ccd9cf46e9be9b3bddb30c3a18892db07/src/signers/seed-signer-btc.js#L80) adds signPsbt; [ISignerEvm](https://github.com/claudiovb/wdk-wallet-evm/blob/e615962e1c081357f45dc58167a94c29cf2d2ae7/src/signers/seed-signer-evm.js#L34) adds signTransaction + signTypedData. The base stays minimal so each chain owns its own contract.

On the "put it on the base with unknown and let subclasses override" alternative ( in despite of the naming conventions mentioned above)...  That just shifts the lying upward: every consumer of ISigner sees a method they can't actually call without already knowing which concrete chain they have, and runtime type checks replace static safety. The current shape also mirrors what IWalletAccount already does (no signTransaction on the read-only base; chain-specific accounts add it), so the asymmetry between the two interfaces stays consistent.

src/signer.js
claudiovb: Which static factory methods do you mean? The ones I'm aware of are things like [WalletAccountBtc.fromPrivateKey()](https://github.com/claudiovb/wdk-wallet-btc/blob/f10be26ccd9cf46e9be9b3bddb30c3a18892db07/src/wallet-account-btc.js#L125), which is itself implemented on top of a PrivateKeySignerBtc... so removing private-key signers would also remove the factories. Is there a different design you have in mind that I'm missing?

Also, even granting that factories cover the single-account case, they return a WalletAccount, not something composable into a WalletManager. If we want the manager to ever accept an imported key (e.g. seed + imported corporate key in the same manager, or seed + hardware wallet), we'd still need ISigner to accommodate non-HD signers unless the manager-accepts-signers concept also goes away, which is the other thread ([#3251895830](https://github.com/tetherto/wdk-wallet/pull/23#discussion_r3251895830)).
But this second point above is a deeper design question. Would need @jonathunne's input on it. Just flagging it here so the linkage is explicit and we don't miss it: if we end up keeping the multi-signer `WalletManager`, then `PrivateKeySignerBtc` probably needs to be a first-class `ISigner` implementation, not an internal-only class hidden from the public contract.

src/signer.js
Davi0kProgramsThings: c.c. @jonathunne 

There is nothing wrong with adding a 'signTransaction(tx: Transaction): unknown' method to the 'ISigner' interface. Then, each signer implementation would still be free to include any other blockchain-specific methods. This way, client code would know that they can use this method to sign transactions. Also, a signer interface that doesn't provide a method to sign transactions would be quite weird.

> On the "put it on the base with unknown and let subclasses override" alternative ( in despite of the naming conventions mentioned above)... That just shifts the lying upward: every consumer of ISigner sees a method they can't actually call without already knowing which concrete chain they have, and runtime type checks replace static safety.

This is not really an issue.

> The current shape also mirrors what IWalletAccount already does (no signTransaction on the read-only base; chain-specific accounts add it)

Actually, the 'IWalletAccount' interface includes the 'signTransaction' method and its return type is 'unknown' as well:
https://github.com/tetherto/wdk-wallet/blob/main/src/wallet-account.js#L77

src/wallet-manager.js
Davi0kProgramsThings: This doesn't check for blank strings: ' ' doesn't throw here.

src/signer.js
Davi0kProgramsThings: While it makes sense i'm wondering if it could actually be useful. It's probably better to exclude it from the common interface and let signer implementations choose how they want to handle this, without enforing it at the super-type level. Implementations could want to use the parent signer configuration for all of its child, or maybe allow overriding just a set of options. Removing it also gets rid of the annoying 'unknown' type.

src/signer.js
Davi0kProgramsThings: c.c. @jonathunne 

Mhhh... How would the 'derive' method work in a private key signer? I guess it would just throw an error. This means that we need to mark it with @throws in the specification of the interface. In my opinion, it would be better to avoid providing private key signers and implement the new static factory methods e.g., 'fromPrivateKey' so that they create the new wallet accounts by skipping derivation and just setting up the internal state to directly use the given private key. This would let us to avoid the @throws in the 'derive' method of the 'ISigner' interface and also the improper use of signers. About using private key signers in wallet managers, i don't think that would make sense at all. Wallet managers only exist to allow users to easily derive multiple accounts from a given seed or signer. Private key signers can't derive other accounts, so with the current 'WalletManager' abstract class they would be totally useless.

src/signer.js
claudiovb: My bad, mistyped, I meant signTypedData/signPSBT. That said, I still think signTransaction doesn't belong on the base: signPSBT isn't just a naming preference, it's the BIP-174 standard and what every major Bitcoin library exposes (bitcoinjs-lib, Ledger, BDK, etc.). IMO we should follow industry standards per chain rather than enforce an internal convention across them. But I'll defer to @jonathunne on the final call.

src/wallet-manager.js
claudiovb: Good catch, I ended up just testing the empty case... tightened the check to cover blank/whitespace-only and non-string inputs (typeof signerName !== 'string' || signerName.trim() === ''). JSDoc updated and added two extra test cases for blank and non-string names.

src/signer.js
claudiovb: Agreed, dropping config keeps the contract clear. Same minimal-base principle I'd argue applies to the sign methods too: ISigner should only promise what's universally applicable across every chain, and anything chain-shaped (config, signPsbt, signTransaction, signTypedData) belongs on the chain-specific extension (ISignerBtc, ISignerEvm). SeedSignerBtc already declares its own typed config on the derive override, so nothing changes at runtime. Pushed the fix, good call.

src/wallet-manager.js
```diff
@@ -91,34 +119,79 @@ export default class WalletManager {
...
   }
 
+  /**
+   * Registers a signer under the given name.
+   *
+   * @param {string} signerName - The signer name.
+   * @param {ISigner} signer - The signer.
+   * @throws {Error} If `signerName` is not a non-empty, non-blank string.
+   */
+  addSigner (signerName, signer) {
```
Davi0kProgramsThings, line 139: This method should return the wallet manager instance in order to allow pipelining operations (as we do in the 'WDK' class):
```suggestion
  /**
   * Registers a signer under the given name.
   *
   * @param {string} signerName - The signer name.
   * @param {ISigner} signer - The signer.
   * @returns {WalletManager} The wallet manager.
   * @throws {Error} If `signerName` is not a non-empty, non-blank string.
   */
  addSigner (signerName, signer) {
```

Davi0kProgramsThings (changes requested):

src/signer.js
Davi0kProgramsThings: In my opinion 'signTransaction' is a perfectly fine name, and it also fits with the other methods of our wallet modules i.e., 'sendTransaction', 'quoteSendTransaction', ... Signer implementations would still be free to provide aliases if they want more blockchain-specific names. Also, 'signTransaction' is specifically for signing transactions. Signers can still provide other blockchain-specific methods e.g., 'signTypedData', 'signPSBT'. Also, the 'unknown' type allows implementations to return what best suits their blockchain. It's still very important for an 'ISigner' interface to provide the specification for a sign method of some sort.

Also once we will start integrating universal signers in our wallet modules, this will make implementing the 'WalletAccount*#signTransaction' method way easier since it will just need to redirect calls to the method above in order to work with all signers.

src/signer.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @interface
+ */
+export class ISigner {
+  /**
+   * Derive a child signer using a relative path (e.g., "0'/0/0").
+   *
+   * @param {string} relPath - The relative derivation path.
+   * @returns {Promise<ISigner>} The derived signer.
+   * @throws {Error} If the signer does not support derivation.
```
Davi0kProgramsThings: Define the following error class in /src/errors.js, and use it:
```javascript
export class SignerError extends Error {
  /**
   * Create a new signer error.
   *
   * @param {string} message - The error's message.
   */
  constructor (message) {
    super(message)

    this.name = 'SignerError'
  }
}
```

src/wallet-manager.js
```diff
@@ -92,34 +120,103 @@ export default class WalletManager {
...
     return this._seed
   }
 
+  /**
+   * Registers a signer under the given name.
+   *
+   * @param {string} signerName - The signer name.
+   * @param {ISigner} signer - The signer.
+   * @returns {WalletManager} The wallet manager.
+   * @throws {Error} If `signerName` is not a non-empty, non-blank string.
```
Davi0kProgramsThings: Note that this should keep throwing 'Error' and not 'SignerError'.

src/wallet-manager.js
```diff
@@ -92,34 +120,103 @@ export default class WalletManager {
...
+   */
+  addSigner (signerName, signer) {
+    if (typeof signerName !== 'string' || signerName.trim() === '') {
```
Davi0kProgramsThings: No need to check types. This check can just be:
```suggestion
    if (!signerName.trim()) {
```

src/wallet-manager.js
```diff
@@ -92,34 +120,103 @@ export default class WalletManager {
...
+      throw new Error('Signer name is required.')
```
Davi0kProgramsThings: Change this to:
```suggestion
      throw new Error('The signer name cannot be an empty or blank string.')
```

src/wallet-manager.js
```diff
@@ -92,34 +120,103 @@ export default class WalletManager {
...
   /**
    * Returns the wallet account at a specific index (see [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)).
    *
-   * @abstract
+   * @overload
    * @param {number} [index] - The index of the account to get (default: 0).
+   * @param {Object} [options] - Account options.
+   * @param {string} [options.signerName] - The signer name. Omit to use the default signer.
    * @returns {Promise<IWalletAccount>} The account.
+   * @throws {Error} If a signer name is given but no signer exists with that name or signer don't support derivation and index is number
```
Davi0kProgramsThings: This method should throw two different errors:
```suggestion
   * @throws {Error} If a signer name is given but no signer exists with that name.
   * @throws {SignerError} If the signer doesn't support account derivation.
```

Davi0kProgramsThings: There's no need to include "index is number" in the error condition since the error itself will only appear when using the method overload that throws it.

src/wallet-manager.js
```diff
@@ -92,34 +120,103 @@ export default class WalletManager {
...
    */
-  async getAccount (index = 0) {
-    throw new NotImplementedError('getAccount(index)')
+
+  /**
+   * Returns the wallet account for a non-derivable signer (e.g., a private-key signer)
```
Davi0kProgramsThings: We should improve the documentation of the method in order to make it clear that it also works for signers that support account derivation. In fact, in such cases it will return the wallet account at the signer's root.

src/wallet-manager.js
```diff
@@ -92,34 +120,103 @@ export default class WalletManager {
...
 
   /**
    * Returns the wallet account at a specific [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) derivation path.
    *
    * @abstract
    * @param {string} path - The derivation path (e.g. "0'/0/0").
+   * @param {Object} [options] - Account options.
+   * @param {string} [options.signerName] - The signer name. Omit to use the default signer.
    * @returns {Promise<IWalletAccount>} The account.
+   * @throws {Error} If a signer name is given but no signer exists with that name, or if the signer does not support derivation.
```
Davi0kProgramsThings: This method should throw two different errors:
```suggestion
   * @throws {Error} If a signer name is given but no signer exists with that name.
   * @throws {SignerError} If the signer doesn't support account derivation.
```

src/wallet-manager.js
```diff
@@ -133,15 +230,25 @@ export default class WalletManager {
...
     for (const account of Object.values(this._accounts)) {
-      if (account.keyPair.privateKey) {
+      if (account.keyPair?.privateKey) {
         account.dispose()
       }
     }
 
+    if (this._defaultSigner !== undefined) {
+      this._defaultSigner.dispose()
+    }
```
Davi0kProgramsThings: This can just be:
```suggestion
    this._defaultSigner?.dispose()
```

tests/wallet-manager.test.js
```diff
@@ -107,4 +131,65 @@ describe('WalletManager', () => {
...
+        .toThrow('Signer name is required.')
+    })
+
+    test('should throw when the signer name is not a string', () => {
+      const signer = new DummySigner()
+      const wallet = new DummyWalletManager(SEED_PHRASE)
+
+      expect(() => wallet.addSigner(null, signer))
+        .toThrow('Signer name is required.')
+    })
```
Davi0kProgramsThings: No need to cover this test case.

types/index.d.ts
Davi0kProgramsThings, line 1: Make sure to update the .d.ts type definitions after addressing the above comments.

Davi0kProgramsThings (changes requested):

src/errors.js
```diff
@@ -25,3 +25,16 @@ export class NotImplementedError extends Error {
     this.name = 'NotImplementedError'
   }
 }
+
+export class SignerError extends Error {
```
Davi0kProgramsThings, line 29: You forgot to export the new error out of the module.

Davi0kProgramsThings (changes requested):

jonathunne (approved):

Davi0kProgramsThings (changes requested):
LGTM.

Davi0kProgramsThings (approved):

PR #24, opened 2025-11-21
Add fiat protocol

nulllpc:
# Description
Add fiat protocol interface to support on/off-ramp protocols


src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * @interface
+ */
+export class IFiatProtocol {
+  /**
+   * Generates a widget URL for a user to purchase a crypto asset with fiat currency.
+   * @param {string} cryptoAsset - The provider-specific code of the crypto asset to purchase.
+   * @param {string} fiatCurrency - The currency's ISO 4217 code (e.g., 'USD').
+   * @param {number} amount - The amount of crypto asset to buy, in its main unit (e.g., 1.50 for 1.50 ETH).
```
Davi0kProgramsThings: This amount should definetely be in base unit rather than main unit in order to maintain consistency across all wallet and protocol methods and avoid precision errors. The type should also accept big integers (as all other token amount arguments do).

```suggestion
   * @param {number | bigint} amount - The amount of crypto asset to buy, in its base unit (e.g., 1,500,000,000,000,000,000 for 1.50 ETH).
```

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,217 @@
...
+   * @param {string} [recipient] - The wallet address to receive the purchased crypto asset.
+   * @param {Record<string, any>} [config] - Provider-specific configuration for the buy operation.
+   * @returns {Promise<string>} The URL for the user to complete the purchase.
+   */
+  async buy (cryptoAsset, fiatCurrency, amount, recipient, config) {
+    throw new NotImplementedError('buy(cryptoAsset, fiatCurrency, amount, recipient, config)')
+  }
```
Davi0kProgramsThings, line 146: Instead of accepting many different positional arguments, this method should accept a single object of type 'BuyOptions'. This would be more consistent with how other methods e.g., transfer, swap, bridge, ... currently work. 

```suggestion
  // In top-level:
  /**
   * @typedef {Object} BuyOptions
   * @property {string} cryptoAsset - The provider-specific code of the crypto asset to purchase.
   * @property {string} fiatCurrency - The currency's ISO 4217 code (e.g., 'USD').
   * @property  {number | bigint} amount - The amount of crypto asset to buy, in its base unit (e.g., 1,500,000,000,000,000,000 for 1.50 ETH).
   * @property {string} [recipient] - The wallet address to receive the purchased crypto asset.
   * @property {Record<string, any>} [config] - Provider-specific configuration for the buy operation.
   */

  /**
   * Generates a widget URL for a user to purchase a crypto asset with fiat currency.
   * @param {BuyOptions} options - The options.
   * @returns {Promise<string>} The URL for the user to complete the purchase.
   */
  async buy (options) {
    throw new NotImplementedError('buy(cryptoAsset, fiatCurrency, amount, recipient, config)')
  }
```

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,217 @@
...
+   * @param {string} cryptoAsset - The provider-specific code of the crypto asset to sell.
+   * @param {string} fiatCurrency - The currency's ISO 4217 code (e.g., 'USD').
+   * @param {number} amount - The amount of crypto asset to sell, in its main unit (e.g., 0.5 for 0.5 ETH).
+   * @param {string} [refundAddress] - The wallet address to receive refunds in case of failure.
+   * @param {Record<string, any>} [config] - Provider-specific configuration for the sell operation.
+   * @returns {Promise<string>} The URL for the user to complete the sale.
+   */
+  async sell (cryptoAsset, fiatCurrency, amount, refundAddress, config) {
+    throw new NotImplementedError('sell(cryptoAsset, fiatCurrency, amount, refundAddress, config)')
+  }
```
Davi0kProgramsThings, line 164: The comments above apply here too.

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,217 @@
...
+  async buy (cryptoAsset, fiatCurrency, amount, recipient, config) {
+    throw new NotImplementedError('buy(cryptoAsset, fiatCurrency, amount, recipient, config)')
+  }
+
+  /**
+   * Generates a widget URL for a user to sell a crypto asset for fiat currency.
+   * @param {string} cryptoAsset - The provider-specific code of the crypto asset to sell.
+   * @param {string} fiatCurrency - The currency's ISO 4217 code (e.g., 'USD').
+   * @param {number} amount - The amount of crypto asset to sell, in its main unit (e.g., 0.5 for 0.5 ETH).
+   * @param {string} [refundAddress] - The wallet address to receive refunds in case of failure.
```
Davi0kProgramsThings: What's the default value of this field? Include it in the description of the argument:
```suggestion
   * @param {string} [refundAddress] - The wallet address to receive refunds in case of failure (default: ?).
```

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,217 @@
...
+    throw new NotImplementedError('sell(cryptoAsset, fiatCurrency, amount, refundAddress, config)')
+  }
+
+  /**
+   * Retrieves the details of a specific transaction from the provider.
+   * @param {'buy' | 'sell'} direction - The direction of the transaction.
+   * @param {string} txId - The unique identifier of the transaction.
+   * @returns {Promise<FiatTransactionDetail>} The transaction details.
+   */
+  async getTransactionDetail (direction, txId) {
```
Davi0kProgramsThings: Isn't the id already enough to fetch the transaction's details?

Davi0kProgramsThings (changes requested):

src/protocols/fiat-protocol.js
nulllpc: i thought about this but unfortunately the amount passed to the widget URL is in main unit. if we pass base unit, then we would need to handle the conversion as well (likely need to pass precision in). that's why i changed `amount` to `number`, i was going for `number | bigint` at first

nulllpc:
technically we don't need to pass an `account` object into `FiatProtocol`, it's only for convenience. do we consider making this optional? or should we prioritize consistency with other modules?

@Davi0kProgramsThings @jonathunne

src/protocols/fiat-protocol.js
nulllpc: unfortunately no, MoonPay has distinct APIs for buy and sell transactions. other protocols have only 1 API endpoint for order detail. personally, i think keeping a `direction` param is simple because it only has 2 possible values and it ensures consistency. it might be redundant in some cases but it's acceptable

src/protocols/fiat-protocol.js
nulllpc: digging deeper into this, MoonPay only understands and returns data according to main unit. here's a metadata of USDC  from MoonPay, USDC has precision of 6 but here it says 2, meaning MoonPay accepts numbers like 0.05

```
{
    "id": "aaefa32f-161b-42c8-8115-debcbf3d6a2d",
    "createdAt": "2019-05-17T18:24:45.206Z",
    "updatedAt": "2019-05-17T18:24:45.206Z",
    "type": "crypto",
    "name": "USD Coin (ERC-20)",
    "code": "usdc",
    "precision": 2,
    "minBuyAmount": 30,
    "maxBuyAmount": 1000,
    "minSellAmount": 15,
    "maxSellAmount": 100000,
    ...
}
```

nulllpc:
one more thing, i took a closer look into various protocols (MoonPay, Ramp, Sardine, Transak) and they might have different definition of a buy/sell:
1. users specify a crypto amount that they want to buy
2. users specify a fiat amount that they want to spend

Moonpay supports both of these cases whereas Sardine only support the 2nd case. I think we should split our `buy` and `sell` into more specific methods:
1. `buyExactCrypto` - current `buy` method
2. `buyWithFiat` - users specify a fiat amount that they want to spend
3. `sellExactCrypto` - current `sell` method
4. `sellForFiat` - users specify a fiat amount they want to receive

what do you guys think? i'll run some more tests to make sure this is accurate and easy-to-use

@Davi0kProgramsThings @jonathunne

src/protocols/fiat-protocol.js
Davi0kProgramsThings: In such case i suggest to get rid of the direction argument in the interface's method, and add it back as a protocol-specific argument in the moon-pay protocol implementation. Otherwise, we might risk to overfit the fiat protocol specification to the moon-pay protocol (while it should be as generic as possible to facilitate its adoption by other protocols).

In order to not violate the liskov substitution principle, the direction argument in the moon-pay protocol should be optional and so it will need a default value.

```javascript
  /**
   * Retrieves the details of a specific transaction from the provider.
   * @override
   * @param {string} txId - The unique identifier of the transaction.
   * @param {'buy' | 'sell'} [direction] - The direction of the transaction (default: 'buy').
   * @returns {Promise<MoonPayTransactionDetail>} The transaction details.
   */
  async getTransactionDetail(txId, direction='buy') {
```

Davi0kProgramsThings:
> technically we don't need to pass an account 'object' into 'FiatProtocol', it's only for convenience. do we consider making this optional? or should we prioritize consistency with other modules?

Other fiat protocols might require it. What you can do here is mark the 'account' argument as optional in the 'FiatProtocol' abstract class constructor and then let subclasses choose whether to make it mandatory, keep it optional or just get rid of it. In the case of the moon-pay protocol, if it doesn't require a wallet account then you can just get rid of it and call the superclass constructor on undefined.

I just took a look at the prototype and i can see that you are actually using it though:
https://github.com/nampc1/wdk-protocol-fiat-moonpay-prototype/blob/refactor-to-official-module/src/moonpay-protocol.js#L302

Davi0kProgramsThings:
> one more thing, i took a closer look into various protocols (MoonPay, Ramp, Sardine, Transak) and they might have different definition of a buy/sell:
> 
> 1. users specify a crypto amount that they want to buy
> 2. users specify a fiat amount that they want to spend

Take a look at how we are currently handling swaps. The way to go here is probably very similar, and it would allow us to keep only two methods: 'buy' and 'sell'. Then protocol implementations can disable one of the two arguments depending on what the protocol actually allows, permitting only option 1. or option 2. operations. 

https://github.com/tetherto/wdk-wallet/blob/main/src/protocols/swap-protocol.js#L28

src/protocols/fiat-protocol.js
jonathunne: Agree with @Davi0kProgramsThings !

jonathunne (changes requested):
Looks great @nampc1 !

Agree with @Davi0kProgramsThings 's comments regarding keeping the `buy` and `sell` methods abstract and leaning on options as arguments to support something like `swapExactTokensForExactTokens` (or `buyExactCrypto`) in this case.

Also - two additions to consider before we ratify this interface - interested in your thoughts:

1. Both `IBridgeProtocol` and `ISwapProtocol` have a config in the constructor that allows the package to be configured with a max fee (`SwapMaxFee` & `BridgeMaxFee`). Is a similar config option worthwhile here? e.g. `FiatMaxFee` 
2. Moonpay supports [retrieving a quote](https://dev.moonpay.com/reference/getbuyquote-1), I suspect other providers also support quoting - seems like a useful primitive to include here - i.e. `quoteBuy` & `quoteSell` methods in the interface.

Davi0kProgramsThings:
1. I'm not sure this would actually fit the case of on/off-ramp protocols. There are no on-chain transactions here, and such protocols usually detain fees directly from the bought or sold amount so allowing users to set a single '*maxFee' option in the constructor wouldn't be enough since operations with a same instance can involve different fiat and crypto currencies. In my opinion, a better solution would be adding some sort of 'minAmountOut' optional argument to the 'buy' and 'sell' methods that basically enables a check that throws if the quote's output amount is less than the argument's value. We are doing something very similar in some of the swap protocols to handle slippage and it works pretty fine. Of course, this approach requires the ability to quote buy and sell operations.
\
If we plan to support any protocols that make use of on-chain transactions in their flow, then we can also think about adding a 'buyMaxFee' and a 'sellMaxFee' optional argument to the 'FiatProtocol' abstract class constructor in the fashion of the '*maxFee' options found in the other protocols (i.e., the max. fee in native coins the user is willing to spend on gas cost for a single buy/sell operation).

    ```javascript
    const url = await protocol.buy({
      [...],
      minAmountOut: 10.0
    })
    ```

src/protocols/fiat-protocol.js
```diff
@@ -62,34 +62,44 @@ import { NotImplementedError } from '../errors.js'
...
  */
 export class IFiatProtocol {
   /**
    * Generates a widget URL for a user to purchase a crypto asset with fiat currency.
-   * @param {string} cryptoAsset - The provider-specific code of the crypto asset to purchase.
-   * @param {string} fiatCurrency - The currency's ISO 4217 code (e.g., 'USD').
-   * @param {number} amount - The amount of crypto asset to buy, in its main unit (e.g., 1.50 for 1.50 ETH).
-   * @param {string} [recipient] - The wallet address to receive the purchased crypto asset.
+   * @param {BuyOptions} options - The options for the buy operation.
    * @param {Record<string, any>} [config] - Provider-specific configuration for the buy operation.
```
Davi0kProgramsThings: In my opinion, the 'config' argument should be part of the 'BuyOptions' and 'SellOptions' types instead of being a separate positional argument. Not only it makes more sense for it to be part of them, but it would also make things easier for protocol implementations that will want to extend the method with extra protocol-specific arguments.

Davi0kProgramsThings (changes requested):

src/protocols/fiat-protocol.js
jonathunne: Yeah, that works too - fee structures likely differ between different providers (just look at how many seperate fees there are in Moonpay), so keeping this as a protocol-specific type is a nice idea. 

FWIW - the reason I suggested adding a config for fees is the fact Moonpay does account for on-chain or "network" fees. It's a returned value when quoting [buys](https://dev.moonpay.com/reference/getbuyquote-1) and [sells](https://dev.moonpay.com/reference/getsellquote-1), or you can [query a dedicated endpoint](https://dev.moonpay.com/reference/getnetworkfees-1) for network fees.

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,221 @@
...
+ */
+
+/**
+ * @interface
+ */
+export class IFiatProtocol {
+  /**
+   * Generates a widget URL for a user to purchase a crypto asset with fiat currency.
+   * @param {BuyOptions} options - The options for the buy operation.
+   * @param {Record<string, any>} [config] - Provider-specific configuration for the buy operation.
```
Davi0kProgramsThings: After a second thought, i think we should just get rid of this argument in the common interface. We don't need a dictionary to allow users to set up arbitrary protocol-specific buy/sell options, as we can just extend the 'BuyOptions'/'SellOptions' type in each protocol implementation with them. e.g., if the moon-pay protocol supports a referral?: string option, we can just override the 'BuyOptions'/'SellOptions' type to add:
```javascript
/**
 * @typedef {Object} BuyOptions
 * @property {string} cryptoAsset - The provider-specific code of the crypto asset to purchase.
 * [...]
 * @property {string} [referral] - The referral code.
 */
```

Davi0kProgramsThings (changes requested):

nulllpc:
Hey @Davi0kProgramsThings, I realized that we can use base units for fiat protocol just like other WDK modules. Basically we can convert into main units when needed using the `decimals` data from protocol's API. I double checked and we can cover most protocols (except for Banxa). When there's no decimals data from the API, we can pass a dictionary in the constructor (applies to Banxa). This way, we would not need to change anything of the current interfaces and `amount` will be consistent with other WDK modules. 

@jonathunne  also approved this.

jonathunne (approved):

jonathunne (changes requested):
@nampc1 , can you include a version bump in this PR?

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,291 @@
...
+ * @typedef {'in_progress' | 'failed' | 'completed'} FiatTransactionStatus
+ */
+
+/**
+ * A protocol-agnostic, standardized object representing the details of an on/off-ramp transaction.
+ * @typedef {Object} FiatTransactionDetail
+ * @property {FiatTransactionStatus} status - The current status of the transaction.
+ * @property {string} cryptoAsset - The provider-specific code of the crypto asset (e.g., 'btc').
+ * @property {string} fiatCurrency - The currency's ISO 4217 code (e.g., 'USD').
+ * @property {Record<string, unknown>} [metadata] - Provider-specific raw data for this transaction.
```
Davi0kProgramsThings: I don't think we really need this field. Fiat protocol implementations can just override the 'FiatTransactionDetail' type and extend it with any protocol-specific fields. Even in the case of a metadata field, overriding is still the best option since it allows protocols to use a better type than Record<string, unknown>. The same applies to all the other types as well (i.e., SupportedCryptoAsset, SupportedFiatCurrency, SupportedCountry, FiatQuote).

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,291 @@
...
+   * @returns {Promise<FiatQuote>} A quote for the transaction.
+   */
+  async quoteBuy (options) {
+    throw new NotImplementedError('quoteBuy(options)')
+  }
+
+  /**
+   * Generates a widget URL for a user to purchase a crypto asset with fiat currency.
+   * @param {BuyOptions} options - The options for the buy operation.
+   * @returns {Promise<string>} The URL for the user to complete the purchase.
```
Davi0kProgramsThings: I would change the return type of the buy and sell methods to FiatQuote & BuyResult. This would be more consistent with the other wallet and protocol modules, where the return types of the methods always include all the quote fields.

If fetching the quote would make the method run much slower in most fiat protocols, then we can avoid doing so.

```suggestion
   * @returns {Promise<FiatQuote & BuyResult>} The operation's result.
```

Davi0kProgramsThings: Instead of returning the url as a raw string, we should encapsulate it in a 'BuyResult' type:

```javascript
/**
 * @typedef {Object} BuyOptions
 * @property {string} url - The url for the user to complete the purchase.
 */
```

```suggestion
   * @returns {Promise<BuyResult>} The operation's result.
```

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,291 @@
...
+   * @returns {Promise<FiatQuote>} A quote for the transaction.
+   */
+  async quoteSell (options) {
+    throw new NotImplementedError('quoteSell(options)')
+  }
+
+  /**
+   * Generates a widget URL for a user to sell a crypto asset for fiat currency.
+   * @param {SellOptions} options - The options for the sell operation.
+   * @returns {Promise<string>} The URL for the user to complete the sale.
```
Davi0kProgramsThings: Instead of returning the url as a raw string, we should encapsulate it in a 'SellResult' type:

```javascript
/**
 * @typedef {Object} SellResult
 * @property {string} url - The url for the user to complete the sale.
 */
```

```suggestion
   * @returns {Promise<SellResult>} The operation's result.
```

src/protocols/index.js
```diff
@@ -30,8 +30,25 @@
 /** @typedef {import('./lending-protocol.js').RepayOptions} RepayOptions */
 /** @typedef {import('./lending-protocol.js').RepayResult} RepayResult */
 
+/** @typedef {import('./fiat-protocol.js').BuyOptions} BuyOptions */
+/** @typedef {import('./fiat-protocol.js').BuyCommonOptions} BuyCommonOptions */
+/** @typedef {import('./fiat-protocol.js').BuyExactCryptoAmountOptions} BuyExactCryptoAmountOptions */
+/** @typedef {import('./fiat-protocol.js').BuyWithFiatAmountOptions} BuyWithFiatAmountOptions */
```
Davi0kProgramsThings, line 36: You can export only 'BuyOptions':
```suggestion
/** @typedef {import('./fiat-protocol.js').BuyOptions} BuyOptions */
```

src/protocols/index.js
```diff
@@ -30,8 +30,25 @@
...
+/** @typedef {import('./fiat-protocol.js').SellOptions} SellOptions */
+/** @typedef {import('./fiat-protocol.js').SellCommonOptions} SellCommonOptions */
+/** @typedef {import('./fiat-protocol.js').SellExactCryptoAmountOptions} SellExactCryptoAmountOptions */
+/** @typedef {import('./fiat-protocol.js').SellForFiatAmountOptions} SellForFiatAmountOptions */
```
Davi0kProgramsThings, line 42: You can export only 'SellOptions':
```suggestion
/** @typedef {import('./fiat-protocol.js').SellOptions} SellOptions */
```

Davi0kProgramsThings (changes requested):
Remember to re-generate the .d.ts type definitions after addressing my comments.

src/protocols/fiat-protocol.js
nulllpc: My rationale is that devs can be aware that there's a `metadata` field, what the `metadata` contain is taken care of by concrete implementation of protocols. Does it make sense to you?

src/protocols/index.js
nulllpc: https://github.com/nampc1/wdk-protocol-fiat-moonpay-fork/blob/main/src/moonpay-protocol.js#L332

Yes, I thought the same at first, but for maximum flexibility, I think we can still export those types. Please see the above example from MoonPay, it allows for selling with crypto amount only, not fiat amount.

Davi0kProgramsThings: Just saw that while i was reviewing the repo. It makes sense, we can keep exporting the partial types as well.

src/protocols/index.js
```diff
@@ -30,8 +30,25 @@
...
+/** @typedef {import('./fiat-protocol.js').FiatQuote} FiatQuote */
+/** @typedef {import('./fiat-protocol.js').FiatTransactionStatus} FiatTransactionStatus */
+/** @typedef {import('./fiat-protocol.js').FiatTransactionDetail} FiatTransactionDetail */
+/** @typedef {import('./fiat-protocol.js').SupportedCryptoAsset} SupportedCryptoAsset */
+/** @typedef {import('./fiat-protocol.js').SupportedFiatCurrency} SupportedFiatCurrency */
+/** @typedef {import('./fiat-protocol.js').SupportedCountry} SupportedCountry */
```
Davi0kProgramsThings, line 52: Give the exports some indentation to make them more readable:
```suggestion
/** @typedef {import('./fiat-protocol.js').BuyOptions} BuyOptions */
/** @typedef {import('./fiat-protocol.js').BuyCommonOptions} BuyCommonOptions */
/** @typedef {import('./fiat-protocol.js').BuyExactCryptoAmountOptions} BuyExactCryptoAmountOptions */
/** @typedef {import('./fiat-protocol.js').BuyWithFiatAmountOptions} BuyWithFiatAmountOptions */

/** @typedef {import('./fiat-protocol.js').SellOptions} SellOptions */
/** @typedef {import('./fiat-protocol.js').SellCommonOptions} SellCommonOptions */
/** @typedef {import('./fiat-protocol.js').SellExactCryptoAmountOptions} SellExactCryptoAmountOptions */
/** @typedef {import('./fiat-protocol.js').SellForFiatAmountOptions} SellForFiatAmountOptions */

/** @typedef {import('./fiat-protocol.js').FiatTransactionDetail} FiatTransactionDetail */
/** @typedef {import('./fiat-protocol.js').FiatTransactionStatus} FiatTransactionStatus */

/** @typedef {import('./fiat-protocol.js').SupportedCryptoAsset} SupportedCryptoAsset */
/** @typedef {import('./fiat-protocol.js').SupportedFiatCurrency} SupportedFiatCurrency */
/** @typedef {import('./fiat-protocol.js').SupportedCountry} SupportedCountry */

/** @typedef {import('./fiat-protocol.js').FiatQuote} FiatQuote */
```

Davi0kProgramsThings (changes requested):

src/protocols/fiat-protocol.js
Davi0kProgramsThings: Knowing that a metadata?: Record<string, unknown> field exists doesn't help developers in any way since they aren't able to do any operations on it as its type is not fully known. A metadata property should also only be used for actual meta data fields, and not for all protocol-specific properties like you are doing in wdk-protocol-fiat-moonpay (which makes accessing those fields unnecessarily more verbose). For these reasons, it's just better to remove the metadata field at all from the common types.

nulllpc: This makes sense, thanks @Davi0kProgramsThings

src/protocols/fiat-protocol.js
nulllpc: We can return the quote in buy/sell methods without affecting performance.

nulllpc: After reviewing MoonPay again, quoting for sell only supports `cryptoAmount`, not `fiatAmount`. however, `sell` supports both `cryptoAmount` and `fiatAmount`. I recommend not calling quote from `buy` and `sell`, this ensures everything still works, developers can call `quoteSell` and `sell` separately

Davi0kProgramsThings (approved):
LGTM.

src/protocols/fiat-protocol.js
```diff
@@ -0,0 +1,296 @@
...
+   * @param {IWalletAccountReadOnly} [account] - The wallet account to use to interact with the protocol.
+   */
+
+  /**
+   * Creates a new fiat protocol with read-only account.
+   *
+   * @overload
+   * @param {IWalletAccount} [account] - The wallet account to use to interact with the protocol.
+   */
+  constructor (account = undefined) {
```
Davi0kProgramsThings: This is slightly better:
```suggestion
  /**
   * Creates a new interface to the moon-pay protocol without binding it to a wallet account.
   *
   * @overload
   * @param {undefined} [account] - The wallet account to use to interact with the protocol.
   */

  /**
   * Creates a new read-only interface to the moon-pay protocol.
   *
   * @overload
   * @param {IWalletAccountReadOnly} account - The wallet account to use to interact with the protocol.
   */

  /**
   * Creates a new interface to the moon-pay protocol.
   *
   * @overload
   * @param {IWalletAccount} account - The wallet account to use to interact with the protocol.
   */
  constructor (account) {
```

Davi0kProgramsThings (changes requested):

src/protocols/fiat-protocol.js
jonathunne: Since this is a generic interface that will support more providers that just Moonpay, can we omit that specific provider in the description, or make it clear Moonpay is an example of a provider that is compatible with IFiatProtocol?

Davi0kProgramsThings: Ohhh, my bad sorry. That was not intended, i just copied and pasted the doc from the moon-pay protocol repo:
```suggestion
  /**
   * Creates a new interface to the protocol without binding it to a wallet account.
   *
   * @overload
   * @param {undefined} [account] - The wallet account to use to interact with the protocol.
   */
  /**
   * Creates a new read-only interface to the protocol.
   *
   * @overload
   * @param {IWalletAccountReadOnly} account - The wallet account to use to interact with the protocol.
   */
  /**
   * Creates a new interface to the protocol.
   *
   * @overload
   * @param {IWalletAccount} account - The wallet account to use to interact with the protocol.
   */
  constructor (account) {
```

jonathunne (approved):

Davi0kProgramsThings (approved):
LGTM.

PR #35, opened 2026-04-17
Feature - Sign tx

Boka44:
# Description
Add a new signTransaction method that signs a transaction and returns the raw signed transaction hex string.

## Motivation and Context
Currently, WDK only supports sendTransaction, which signs and broadcasts in one step. Some consumers want to sign a transaction locally and handle broadcasting themselves (e.g. to submit via a different relayer, queue the tx, or inspect it before sending). This new method supports that workflow while leaving sendTransaction unchanged.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### DIFF
--- a/src/wallet-account.js
+++ b/src/wallet-account.js
@@ -72,7 +72,7 @@ export class IWalletAccount extends IWalletAccountReadOnly {
    * Signs a transaction
    *
    * @param {Transaction} tx - The transaction to sign.
-   * @returns {Promise<string>} The signed transaction as a hex string.
+   * @returns {Promise<unknown>} The signed transaction.
    */
   async signTransaction (tx) {
     throw new NotImplementedError('signTransaction(tx)')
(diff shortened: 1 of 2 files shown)


src/wallet-account.js
```diff
@@ -68,6 +68,16 @@ export class IWalletAccount extends IWalletAccountReadOnly {
     throw new NotImplementedError('sign(message)')
   }
 
+  /**
+   * Signs a transaction
+   *
+   * @param {Transaction} tx - The transaction to sign.
+   * @returns {Promise<unknown>} The signed transaction.
```
Davi0kProgramsThings, line 75: Can't we just always return a string? It should be pretty standard to serialize signatures to hex strings. This way we could avoid using unknown for the return type.

Davi0kProgramsThings (changes requested):

PR #39, opened 2026-05-15
feat: Swidge protocol

jonathunne:
## Summary
Introduces `SwidgeProtocol` (Swidge = Swap + Bridge), a protocol interface for providers that execute a swap and bridge as one cross-chain action.

The interface follows a quote-first flow:
  - `quoteSwidge(options)` returns the expected input/output amounts, minimum received amount, fees, estimated duration, price impact, expiry, and provider-specific data.
  - `executeSwidge(quote, config?)` executes a previously returned quote.
  - `getSwidgeStatus(id, options?)` tracks asynchronous cross-chain execution.
  - `getSupportedChains()` and `getSupportedTokens(options?)` expose provider-supported routes and assets.

## Motivation
Some providers now offer atomic bridge-and-swap flows that do not fit cleanly into the existing separate `SwapProtocol` and `BridgeProtocol` abstractions. In those systems, quote generation, execution data, routing, refunds, and status tracking are tied to one provider-level operation. This interface gives those providers a dedicated abstraction without forcing implementations to split a single provider action across two protocols, or extend `IBridgeProtocol` in a custom provider-specific way.

##  Notes
- The protocol keeps common fields first-class where they are broadly supported: quoted amounts, slippage-adjusted minimum output, fees, estimated duration, quote expiry, price impact, refund address, transactions, and status.
- Provider-specific quote/execution details remain in `providerData` because routing models differ significantly across providers. For example, some providers return ready-to-sign transactions, some use deposit addresses, some require order signatures, and some expose multi-step route metadata.
- Token and chain discovery are included because most researched providers expose supported assets and networks, providers that don't support this can throw "Not Implemented".


src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+    throw new NotImplementedError('getSupportedTokens(options)')
+  }
+}
+
+/**
+ * @abstract
+ * @implements {ISwidgeProtocol}
+ */
+export default class SwidgeProtocol {
+  /**
```
AlonzoRicardo, line 227: `executeSwidge(quote, config)` doesn't have access to `recipient`, `refundAddress`, `slippage`, or `toChain` — won't implementers be forced to smuggle those through `providerData`?

Would `executeSwidge(quote, options, config)` work better? It mirrors how `SwapProtocol.swap(options)` / `BridgeProtocol.bridge(options)` already pass full options at execution time.

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+ * @typedef {Object} SwidgeStatusResult
+ * @property {SwidgeStatus} status - The current status of the swidge.
+ * @property {SwidgeTransaction[]} [transactions] - Transactions associated with the swidge.
+ * @property {object} [providerData] - Opaque provider-specific data.
+ */
+
+/**
+ * @typedef {Object} SwidgeSupportedChain
+ * @property {string | number} id - The provider-specific chain identifier.
+ * @property {string} name - The human-readable chain name.
```
AlonzoRicardo, line 129: For `SwapProtocol`, a scalar `fee` is unambiguous because it's always source-chain gas. A swidge can charge fees across multiple tokens and chains in one operation — what does a single `bigint fee` (or `swidgeMaxFee` cap) actually represent in that case?

E.g. a route with gas on the source chain plus a protocol fee in a stablecoin plus a bridge fee in a third token — three denominations. Curious how you're thinking about this, given `fees[]` already carries the per-token/per-chain detail.

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+ * @property {string} [type] - The chain or virtual machine type (e.g., 'evm', 'svm', 'utxo').
+ * @property {string} [nativeToken] - The symbol of the chain's native token.
+ * @property {object} [providerData] - Opaque provider-specific metadata.
+ */
+
+/**
```
AlonzoRicardo, line 134: `SwidgeQuote.providerData` is required but `SwidgeResult.providerData` / `SwidgeStatusResult.providerData` are optional — intentional? Could be made optional here too for providers that don't need it.

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+ * @property {bigint} [swidgeMaxProtocolFee] - The maximum protocol fee for swidge operations.
+ */
+
+/**
+ * @typedef {SwidgeCommonOptions & (SwidgeExactInOptions | SwidgeExactOutOptions)} SwidgeOptions
+ */
+
+/**
+ * @typedef {Object} SwidgeCommonOptions
+ * @property {string} fromToken - The provider-specific identifier or address of the source token.
```
AlonzoRicardo, line 51: Worth a deliberate decision on naming — Swidge introduces a third vocabulary alongside Swap and Bridge:

| | Swap | Bridge | Swidge |
|---|---|---|---|
| Source token | `tokenIn` | `token` | `fromToken` |
| Destination token | `tokenOut` | — | `toToken` |
| Recipient | `to` | `recipient` | `recipient` |
| Variant | `Sell`/`Buy` | n/a | `ExactIn`/`ExactOut` |

The new names are arguably clearer for cross-chain, but a provider implementing both `ISwapProtocol` and `ISwidgeProtocol` ends up with parallel vocabularies for the same primitive. Align to existing names, or adopt these and plan to converge the others?

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+ * @property {string} symbol - The token symbol.
+ * @property {number} decimals - The number of decimal places for the token's base unit.
+ * @property {string} [address] - The token contract address, if applicable.
+ * @property {string} [name] - The token's full name.
+ * @property {object} [providerData] - Opaque provider-specific metadata.
+ */
+
+/**
+ * @typedef {Object} SwidgeSupportedTokensOptions
+ * @property {string | number} [chain] - Filters tokens by chain.
```
AlonzoRicardo: Three chain filters here — `chain`, `fromChain`, `toChain`. When would a consumer use `chain` vs `fromChain`? Could this collapse to just `fromChain` / `toChain`?

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+   */
+
+  /**
+   * Creates a new swidge protocol.
+   *
+   * @overload
+   * @param {IWalletAccount} account - The wallet account to use to interact with the protocol.
+   * @param {SwidgeProtocolConfig} [config] - The swidge protocol configuration.
+   */
+  constructor (account, config = {}) {
```
AlonzoRicardo, line 242: `getSupportedChains()` and `getSupportedTokens()` don't actually need a wallet account to execute. `FiatProtocol` allows `account = undefined` for exactly this case — worth mirroring so consumers can do discovery (e.g. populate a chain dropdown) before a wallet is connected?

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+ * @property {object} [providerData] - Opaque provider-specific data to assist status lookup.
+ */
+
+/**
+ * @typedef {Object} SwidgeStatusResult
+ * @property {SwidgeStatus} status - The current status of the swidge.
+ * @property {SwidgeTransaction[]} [transactions] - Transactions associated with the swidge.
+ * @property {object} [providerData] - Opaque provider-specific data.
+ */
+
```
AlonzoRicardo, line 125: Multi-step routing was deferred to phase 2, but worth flagging the concrete case: LiFi (one of the providers we're targeting) exposes routes as `Route.steps[]` — a single cross-chain action like USDC on Polygon → ARB on Arbitrum can decompose into swap → bridge → swap, three sequential txs the consumer has to sign in order. Squid Router has the same shape.

If LiFi gets implemented against this interface later, consumers built around "one quote → one execute" will need to restructure. Worth a JSDoc note now stating `SwidgeQuote` represents a single-step route so consumers don't over-fit?

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  /**
+   * Executes a previously quoted swidge.
+   *
+   * @param {SwidgeQuote} quote - The quote returned by quoteSwidge.
+   * @param {SwidgeProtocolConfig} [config] - Optional fee limits for the execution.
+   * @returns {Promise<SwidgeResult>} The swidge execution result.
+   */
+  async executeSwidge (quote, config) {
```
Davi0kProgramsThings: In order to make the method more consistent with the other protocols, i would rename it to just:
```suggestion
  async swidge (quote, config) {
```

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+   * @returns {Promise<SwidgeQuote>} The quoted swidge details.
+   */
+  async quoteSwidge (options) {
+    throw new NotImplementedError('quoteSwidge(options)')
+  }
+
+  /**
+   * Executes a previously quoted swidge.
+   *
+   * @param {SwidgeQuote} quote - The quote returned by quoteSwidge.
```
Davi0kProgramsThings: This is quite different with how the other protocols work. Normally, users are able to execute a swap or a bridge even without quoting it first. Is there a specific reason we should change it in this new protocol? I guess it's to avoid extra RPC calls to fetch the operation quote, if that's the case we could easily fix it internally by implementing a simple cache just like we are doing for the new version of wallet-evm-erc-4337 that uses abstraction kit.

```suggestion
   * @param {SwidgeOptions} options - The swidge options.
```

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+  /**
+   * Retrieves the current status of an in-flight swidge.
+   *
+   * @param {string} id - The swidge execution identifier returned by executeSwidge.
+   * @param {SwidgeStatusOptions} [options] - Optional hints to assist provider lookups.
+   * @returns {Promise<SwidgeStatusResult>} The current swidge status.
+   */
+  async getSwidgeStatus (id, options) {
+    throw new NotImplementedError('getSwidgeStatus(id, options)')
+  }
```
Davi0kProgramsThings, line 190: What should happen if the id is invalid or doesn't exist? We can either return null or throw an error. Here, the latter option is probably slightly better so i would go with it:
```suggestion
  /**
   * Retrieves the current status of an in-flight swidge.
   *
   * @param {string} id - The swidge execution identifier returned by executeSwidge.
   * @param {SwidgeStatusOptions} [options] - Optional hints to assist provider lookups.
   * @returns {Promise<SwidgeStatusResult>} The current swidge status.
   * @throws {Error} If the id is invalid, or no order exists with the given identifier.
   */
  async getSwidgeStatus (id, options) {
    throw new NotImplementedError('getSwidgeStatus(id, options)')
  }
```

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+  async getSupportedTokens (options) {
+    throw new NotImplementedError('getSupportedTokens(options)')
+  }
+}
+
+/**
+ * @abstract
+ * @implements {ISwidgeProtocol}
+ */
+export default class SwidgeProtocol {
```
Davi0kProgramsThings, line 218: We should be able to do something very powerful here, which is making the 'SwidgeProtocol' class implement the 'ISwapProtocol' and 'IBridgeProtocol' interfaces as well. We can provide a concrete implementation for the swap and bridge methods that just adapts their arguments to the 'SwidgeOptions' type and then calls this.swidge with the proper parameters to execute the given swap or bridge operation. Basically, we would make all instances of the 'SwidgeProtocol' class valid swap and bridge protocols without requiring any action on the developers or the client code. This makes sense since an operation that swaps a pair of tokens on a same blockchain is in fact a valid swap operation, and an operation that bridges a single token between two different blockchains is in fact a valid bridge operation.

src/protocols/swidge-protocol.js
jonathunne: Good catch @AlonzoRicardo  - in favour of adopting these and converging `ISwapProtocol` & `IBridgeProtocol`.

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+
+/**
+ * @typedef {'gas' | 'protocol' | 'bridge' | 'relayer'
+ *          | 'application' | 'affiliate' | 'liquidity' | 'other'} SwidgeFeeType
+ */
+
+/**
+ * @typedef {Object} SwidgeProtocolConfig
+ * @property {bigint} [swidgeMaxFee] - The maximum total fee for swidge operations.
+ * @property {bigint} [swidgeMaxProtocolFee] - The maximum protocol fee for swidge operations.
```
Davi0kProgramsThings: These two options should also allow numbers:
```suggestion
 * @property {number | bigint} [swidgeMaxFee] - The maximum total fee for swidge operations.
 * @property {number | bigint} [swidgeMaxProtocolFee] - The maximum protocol fee for swidge operations.
```

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+ * @property {object} providerData - Opaque provider-specific data required to execute the swidge.
+ */
+
+/**
+ * @typedef {Object} SwidgeResult
+ * @property {string} id - The unique swidge execution identifier.
+ * @property {string} [hash] - The primary transaction hash (if available immediately).
+ * @property {bigint} [fee] - The total aggregated fee charged.
+ * @property {bigint} [protocolFee] - The protocol-specific fee charged.
+ * @property {SwidgeFee[]} [fees] - Itemized fee breakdown.
```
Davi0kProgramsThings: Is there a reason these fields are all optional? I guess it's to accomodate different protocols that don't share a common interface. This would not be the proper way to do it though, we should select a list of properties that all protocols can set and use optional only for fields which valid values actually include undefined (and not to indicate a field that could or couldn't exist depending on the implementation).

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+ * @property {bigint} fromTokenAmount - The amount of source tokens to spend.
+ * @property {bigint} toTokenAmount - The estimated amount of destination tokens to receive.
+ * @property {bigint} toTokenAmountMin - The minimum guaranteed amount after slippage.
+ * @property {bigint} [fee] - The total aggregated fee.
+ * @property {bigint} [protocolFee] - The protocol-specific fee portion.
+ * @property {SwidgeFee[]} [fees] - Itemized fee breakdown.
+ * @property {number} [estimatedDuration] - Estimated duration in seconds.
+ * @property {number} [expiry] - Unix timestamp (seconds) at which the quote expires.
+ * @property {number} [priceImpact] - Provider-reported estimated price impact as a decimal (e.g., 0.01 for 1%).
+ * @property {object} providerData - Opaque provider-specific data required to execute the swidge.
```
Davi0kProgramsThings: Instead of object, prefer:
```suggestion
 * @property {Record<string, unknown>} providerData - Opaque provider-specific data required to execute the swidge.
```

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,298 @@
...
+ */
+
+/**
+ * @typedef {Object} SwidgeQuote
+ * @property {bigint} fromTokenAmount - The amount of source tokens to spend.
+ * @property {bigint} toTokenAmount - The estimated amount of destination tokens to receive.
+ * @property {bigint} toTokenAmountMin - The minimum guaranteed amount after slippage.
+ * @property {bigint} [fee] - The total aggregated fee.
+ * @property {bigint} [protocolFee] - The protocol-specific fee portion.
+ * @property {SwidgeFee[]} [fees] - Itemized fee breakdown.
```
Davi0kProgramsThings: Does it make sense for a quote to not include these fields? I'm not sure why they should be optional, especially 'fee'. Returning an estimation of the cost of an operation is basically the main purpose of the quote methods.

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,310 @@
...
+ * @property {number} [estimatedDuration] - Estimated duration in seconds.
+ * @property {number} [expiry] - Unix timestamp (seconds) at which the quote expires.
+ * @property {number} [priceImpact] - Provider-reported estimated price impact as a decimal (e.g., 0.01 for 1%).
+ * @property {object} [providerData] - Opaque provider-specific data required to execute the swidge.
```
Davi0kProgramsThings: This is not the proper way to allow protocols to accept or return new properties. Instead, protocol implementations should extend the 'SwidgeQuote' type with any protocol-specific fields by typing and documenting them properly. For this reason we should probably get rid of the provider data field from this type and from all the others below. Generally, if the type that better suits a property is object in most cases it means that the field shouldn't be there. The only exception is if the field can actually contain arbitrary data chosen by the user (not the protocol).

Davi0kProgramsThings (changes requested):

src/protocols/swidge-protocol.js
Davi0kProgramsThings: We can cache and fetch the quote internally, in my opinion the 'swidge' method should just accept 'options' and 'config' just like the other protocols. I left a comment about this in my code review by the way.

src/protocols/swidge-protocol.js
jonathunne: Agree - done!

src/protocols/swidge-protocol.js
jonathunne: Agree

src/protocols/swidge-protocol.js
jonathunne: Yeah, prefer that - done

src/protocols/swidge-protocol.js
jonathunne: Thanks guys - replying here, but relevant for a few comments further down.

Original reasoning was that swidge providers tend to issue stateful & time-sensitive quotes (1inch Fusion quoteId, LiFi route IDs, Aori RFQs, etc.), so the idea was for the quote to be a binding artifact so consumers knew exactly which quote was executing. Agree with @Davi0kProgramsThings that an internal cache covers stateful providers, plus we can throw `QuoteExpiredError` or similar on stale quotes where applicable.

Happy to move to `swidge(options, config)` as suggested, and having `ISwidgeProtocol` also implement `ISwapProtocol` and `IBridgeProtocol`, but we should keep in mind:

1. Not every swidge provider can satisfy both contracts. e.g. Chainflip is purely cross-chain (no same-chain swap). I think this is acceptable, throwing "Not Supported" errors is a convention in WDK.
2. `SwapResult.hash` is required but `SwidgeResult.hash` is optional (Chainflip returns a deposit channel ID first, hash later). `SwapResult.fee` / `BridgeResult.fee+bridgeFee` are `bigint`, but `SwidgeResult.fees` is multi-denomination. Need to reconcile these.


I think (2) is the biggest blocker, everything else is reasonably straightforward.

lokeshagrawal99: But we were planning to deprecate ISwapProtocol and IBridge protocol in future and migrate existing providers to swidge. So if we implement swap and bridge in SwidgeProtocol, we won't be able to deprecate them. So are we changing that philosophy?

Davi0kProgramsThings: 1. This shouldn't really be an issue. If a protocol doesn't support single-chain swaps, then the 'swidge' method should already throw in the case the source and the target blockchains are equal. The implementation of the 'swap' method in the 'SwidgeProtocol' abstract class will take care of mapping and adapting the 'options' argument according to the 'SwidgeOptions' type, which will also include setting the 'toChain' property to the input blockchain. In the case of protocols that only support cross-chain swaps, this will cause the method to throw the proper error without any further changes. 
2. We can use 'hash' for the id, and introduce a new 'transactionHash' field for the actual transaction hash. The 'hash' field is meant to include a unique identifier for the operation, so in terms of specification this would work just fine. The only issue is that the name 'hash' wouldn't really fit the property anymore. We could think about changing it to 'id' across all protocols, but i don't really think it would be worth the breaking change. About 'fee', we can think about returning some sort of accumulator of all the fees e.g., the total fee amount in dollars. Alternatively, we could return the cost of the operation on the source blockchain and then add a separate 'fees: SwidgeFee[]' field to an override of the 'SwapResult' type.

Where does the idea of getting rid of the swap and bridge protocols come from? I don't think it makes much sense. All swap and bridge protocols that don't support cross-operations would need to throw a bunch of errors, which would just bloat the specification of the 'ISwidgeProtocol' interface even if actual swidge protocols would never throw them.

jonathunne: Re. (1) - fully agree.

Re. the id - yeah that's why I went with id in `SwidgeResult`. Happy to keep hash if we clarify in the JSDocs that it doesn't need to be a tx hash.

Re. fees - agree on the override. Not sure about the accumulator idea - a lot of providers don't return USD-denominated fees natively, so this introduces some complexity to each package if we mandate it as a part of the interface.

Re. deprecation (@lokeshagrawal99) - I think I put that as an open question in the rationale doc, not a stated plan. Aligned with @Davi0kProgramsThings here - single-purpose providers (e.g. Velora) shouldn't have to throw errors on cross-chain methods.

Davi0kProgramsThings: It seems that we should be able to proceed with making the 'ISwidgeProtocol' interface extend 'ISwapProtocol' and 'IBridgeProtocol'. I'm not sure if we should make the 'ISwidgeProtocol' interface extend the other two, or just make the 'SwidgeProtocol' abstract class implement all of them. The only difference is that the latter allows developers that don't want to use the abstract class to quickly implement the interface since it wouldn't include the swap and bridge methods. However, why would someone choose to not use the abstract class? For this reason, making the interface extend 'ISwapProtocol' and 'IBridgeProtocol' is probably the best option.

src/protocols/swidge-protocol.js
jonathunne: Thanks, and agree on the principle. After some thought I made `type` and `nativeToken` required on `SwidgeSupportedChain`, since most providers support these, or should be able to infer. `address` on `SwidgeSupportedToken` stays optional (this could genuinely be underfeed for native tokens like ETH).

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,401 @@
...
+
+  /**
+   * Quotes the costs of a bridge operation.
+   *
+   * @param {BridgeOptions} options - The bridge's options.
+   * @returns {Promise<Omit<BridgeResult, 'hash'>>} The bridge's quotes.
+   */
+  async quoteBridge (options) {
+    throw new NotImplementedError('quoteBridge(options)')
+  }
```
Davi0kProgramsThings: No need to include these methods since the interface already inherits them from 'ISwapProtocol' and 'IBridgeProtocol'. In javascript, sub-types are not meant to extend 'ISwidgeProtocol' but even if they do it wouldn't be an issue to not have an implementation for these methods since they always throw. We can make the class and its type definition cleaner and more correct by getting rid of them.

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,401 @@
...
+  /**
+   * Quotes the costs of a bridge operation.
+   *
+   * @abstract
+   * @param {BridgeOptions} options - The bridge's options.
+   * @returns {Promise<Omit<BridgeResult, 'hash'>>} The bridge's quotes.
+   */
+  async quoteBridge (options) {
+    throw new NotImplementedError('quoteBridge(options)')
+  }
```
Davi0kProgramsThings, line 340: > https://github.com/tetherto/wdk-wallet/pull/39/files#r3275515155<br/>We can provide a concrete implementation for the swap and bridge methods that just adapts their arguments to the 'SwidgeOptions' type and then calls this.swidge with the proper parameters to execute the given swap or bridge operation. Basically, we would make all instances of the 'SwidgeProtocol' class valid swap and bridge protocols without requiring any action on the developers or the client code.

The idea above requires an additional step which is giving a concrete implementation to the 'swap' and 'bridge' methods that just wraps a call to 'swidge' by adapting the swap/bridge options to the proper type. This way, sub-types wouldn't need to provide an implementations for those methods which would make the protocol easier to extend and use. The same also applies to the quote methods which should just wrap a call to 'quoteSwidge'. Without this, we would make extending the abstract class very annoying since developers should have to manually provide an implementation to all the swap, bridge and swidge methods. This is an example of what the implementation of the 'swap' method should look like:
```javascript
  /**
   * Swaps a pair of tokens.
   *
   * @param {SwapOptions} options - The swap's options.
   * @returns {Promise<SwapResult>} The swap's result.
   */
  async swap (options) {
    const { id: hash, fees } = await this.swidge({
      fromToken: options.tokenIn,
      toToken: options.tokenOut,
      toChain: *,
      recipient: options.to,
      fromTokenAmount: options.tokenInAmount,
      toTokenAmount: options.tokenOutAmount
    })

    // The fees should all be in the same native coin or token since the swap is not cross-chain:
    const fee = fees.reduce((acc, { amount }) => acc + amount, 0n)

    return { hash, fee }
  }
```

\* 'toChain' is the only field that is causing troubles at the moment, since in the abstract class we don't know what network will be the source blockchain. We have two options to fix this issue:
1. mark the 'toChain' argument as optional in the 'SwidgeOptions' type and make it use the source blockchain by default.
2. require the source blockchain in the configuration object at construction i.e., add a chain id or similar field to the 'SwidgeProtocolConfig' type.

Davi0kProgramsThings (changes requested):

src/protocols/swidge-protocol.js
jonathunne: Thanks @Davi0kProgramsThings , mostly agree. `swap` & `quoteSwap` are unproblematic to implement concrete delegations for, but I'm not sure there is a clean mapping for `bridge` though: `BridgeResult` expects `fee` `bridgeFee` as `bigints`, but `SwidgeFee[]` can span multiple denominations and chains - meaning the delegation might need to sum up amounts to fit the interface. In the absence of a clean alternative, I am in favour of keeping `bridge` & `quoteBridge` as stubs for implementations to override with provider-specific details about fees.

Re. `toChain`, prefer option 1. Also, implementations should know the source chain they can retrieve from the provider on the wallet account or infer it.

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,385 @@
...
+ *          | 'refund-pending' | 'refunded' | 'cancelled' | 'expired' | 'partial'} SwidgeStatus
+ */
+
+/**
+ * @typedef {'gas' | 'protocol' | 'bridge' | 'relayer'
+ *          | 'application' | 'affiliate' | 'liquidity' | 'other'} SwidgeFeeType
+ */
+
+/**
+ * @typedef {Object} SwidgeProtocolConfig
```
AlonzoRicardo, line 40: Curious — `swidgeMaxFee` / `swidgeMaxProtocolFee` got dropped when we moved to `fees[]`. Intentional, or was a `maxFees?: SwidgeFee[]` shape on the table?

src/protocols/swidge-protocol.js
Davi0kProgramsThings: What about?
```javascript
  /**
   * Bridges a token to a different blockchain.
   *
   * @param {BridgeOptions} options - The bridge's options.
   * @returns {Promise<BridgeResult>} The bridge's result.
   */
  async bridge (options) {
    const { id: hash, fees } = await this.swidge({
      fromToken: options.token,
      toToken: options.token,
      toChain: options.targetChain,
      recipient: options.recipient,
      fromTokenAmount: options.amount
    })

    // We should be able to split the fees in two different lists: one that contains the gas cost of the bridge transactions, and one that contains the bridge fees:
    const fee = fees.filter(fee => fee.type == 'gas')
                    .reduce((acc, { amount }) => acc + amount, 0n)

    const bridgeFee = fees.filter(fee => fee.type == 'bridge')
                          .reduce((acc, { amount }) => acc + amount, 0n)

    return { hash, fee, bridgeFee }
  }
```

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,406 @@
...
+      toToken: options.tokenOut,
+      recipient: options.to
+    }
+    if (options.tokenInAmount !== undefined) {
+      swidgeOptions.fromTokenAmount = options.tokenInAmount
+    } else {
+      swidgeOptions.toTokenAmount = options.tokenOutAmount
+    }
+
+    const result = await this.swidge(swidgeOptions)
```
Davi0kProgramsThings: In the 'SwapOptions' type, the 'tokenInAmount' and 'tokenOutAmount' fields are mutually exclusive so you can refactor this code to just:
```suggestion
    const result = await this.swidge({
      fromToken: options.tokenIn,
      toToken: options.tokenOut,
      recipient: options.to,
      fromTokenAmount: options.tokenInAmount,
      toTokenAmount: options.tokenOutAmount
    })
```

src/protocols/swidge-protocol.js
```diff
@@ -0,0 +1,406 @@
...
+      toToken: options.tokenOut,
+      recipient: options.to
+    }
+    if (options.tokenInAmount !== undefined) {
+      swidgeOptions.fromTokenAmount = options.tokenInAmount
+    } else {
+      swidgeOptions.toTokenAmount = options.tokenOutAmount
+    }
+
+    const quote = await this.quoteSwidge(swidgeOptions)
```
Davi0kProgramsThings:
```suggestion
    const result = await this.quoteSwidge({
      fromToken: options.tokenIn,
      toToken: options.tokenOut,
      recipient: options.to,
      fromTokenAmount: options.tokenInAmount,
      toTokenAmount: options.tokenOutAmount
    })
```

types/src/protocols/swidge-protocol.d.ts
```diff
@@ -0,0 +1,518 @@
...
+ * @property {number} decimals - The number of decimal places for the token's base unit.
+ * @property {string} [address] - The token contract address, if applicable.
+ * @property {string} [name] - The token's full name.
+ */
+/**
+ * @typedef {Object} SwidgeSupportedTokensOptions
+ * @property {string | number} [fromChain] - The optional source chain for route-scoped discovery.
+ * @property {string} [fromToken] - The optional source token for route-scoped discovery.
+ * @property {string | number} [toChain] - The optional destination chain for route-scoped discovery.
+ */
```
Davi0kProgramsThings: Get rid of these comments at the beginning of the .d.ts file.

types/src/protocols/swidge-protocol.d.ts
```diff
@@ -0,0 +1,518 @@
...
+     * @param {SwidgeSupportedTokensOptions} [options] - Optional filters for chain- or route-scoped token discovery.
+     * @returns {Promise<SwidgeSupportedToken[]>} The supported tokens.
+     */
+    getSupportedTokens(options?: SwidgeSupportedTokensOptions): Promise<SwidgeSupportedToken[]>;
+}
+/**
+ * @abstract
+ * @implements {ISwidgeProtocol}
+ * @implements {ISwapProtocol}
+ * @implements {IBridgeProtocol}
```
Davi0kProgramsThings: You can get rid of these tags in the type definitions.

types/src/protocols/swidge-protocol.d.ts
```diff
@@ -0,0 +1,518 @@
...
+     */
+    getSupportedChains(): Promise<SwidgeSupportedChain[]>;
+    /**
+     * Retrieves the tokens supported by the provider for swidge operations.
+     *
+     * @abstract
+     * @param {SwidgeSupportedTokensOptions} [options] - Optional filters for chain- or route-scoped token discovery.
+     * @returns {Promise<SwidgeSupportedToken[]>} The supported tokens.
+     */
+    getSupportedTokens(options?: SwidgeSupportedTokensOptions): Promise<SwidgeSupportedToken[]>;
```
Davi0kProgramsThings: For some reason, tsc didn't add the abstract keyword to these methods:
```suggestion
    /**
     * Quotes the estimated costs and output of a swidge operation.
     * Returns a non-binding quote; the actual execution is performed
     * by {@link swidge}.
     *
     * @abstract
     * @param {SwidgeOptions} options - The swidge options.
     * @returns {Promise<SwidgeQuote>} The quoted swidge details.
     */
    abstract quoteSwidge(options: SwidgeOptions): Promise<SwidgeQuote>;
    /**
     * Executes a swidge operation.
     *
     * @abstract
     * @param {SwidgeOptions} options - The swidge options.
     * @param {SwidgeProtocolConfig} [config] - Optional provider-specific execution configuration.
     * @returns {Promise<SwidgeResult>} The swidge execution result.
     */
    abstract swidge(options: SwidgeOptions, config?: SwidgeProtocolConfig): Promise<SwidgeResult>;
    /**
     * Retrieves the current status of an in-flight swidge.
     *
     * @abstract
     * @param {string} id - The swidge execution identifier returned by swidge.
     * @param {SwidgeStatusOptions} [options] - Optional hints to assist provider lookups.
     * @returns {Promise<SwidgeStatusResult>} The current swidge status.
     * @throws {Error} If the id is invalid, or no swidge exists with the given identifier.
     */
    abstract getSwidgeStatus(id: string, options?: SwidgeStatusOptions): Promise<SwidgeStatusResult>;
    /**
     * Retrieves the chains supported by the provider for swidge operations.
     *
     * @abstract
     * @returns {Promise<SwidgeSupportedChain[]>} The supported chains.
     */
    abstract getSupportedChains(): Promise<SwidgeSupportedChain[]>;
    /**
     * Retrieves the tokens supported by the provider for swidge operations.
     *
     * @abstract
     * @param {SwidgeSupportedTokensOptions} [options] - Optional filters for chain- or route-scoped token discovery.
     * @returns {Promise<SwidgeSupportedToken[]>} The supported tokens.
     */
    abstract getSupportedTokens(options?: SwidgeSupportedTokensOptions): Promise<SwidgeSupportedToken[]>;
```

types/src/protocols/swidge-protocol.d.ts
```diff
@@ -0,0 +1,518 @@
...
+}
+/**
+ * @abstract
+ * @implements {ISwidgeProtocol}
+ * @implements {ISwapProtocol}
+ * @implements {IBridgeProtocol}
+ */
+export default class SwidgeProtocol implements ISwidgeProtocol, ISwapProtocol, IBridgeProtocol {
```
Davi0kProgramsThings: Missing abstract keyword:
```suggestion
export default abstract class SwidgeProtocol implements ISwidgeProtocol, ISwapProtocol, IBridgeProtocol {
```

Davi0kProgramsThings (changes requested):

Davi0kProgramsThings (approved):
LGTM.

PR #47, opened 2026-06-26
remove signTransaction from ISigner interface, update deps

claudiovb:
# Description
Removes `signTransaction` from the base `ISigner` interface so concrete signer implementations are free to use their chain-idiomatic signing method (e.g., `signPsbt` for BTC, `signTransaction` for EVM).

## Motivation and Context
Forcing a uniform `signTransaction` name on the base contract was misleading, the operation has genuinely divergent naming across chains, also the input type varies a lot. Keeping `ISigner` minimal and chain-agnostic lets implementers follow each ecosystem's conventions without aliasing.

## Type of change

- [x] Feature removal (removal of a method from a public base interface)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1189,13 +1189,13 @@
       "license": "MIT"
     },
     "node_modules/@types/node": {
-      "version": "25.9.3",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-25.9.3.tgz",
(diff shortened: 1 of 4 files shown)


Davi0kProgramsThings (approved):
LGTM.

AlonzoRicardo (approved):

jonathunne (approved):

PR #49, opened 2026-07-01
fix(types): default TSignedTransaction generic to unknown

AlonzoRicardo:
## Summary

`IWalletAccount<TSignedTransaction>` was made generic in `1.0.0-beta.12` (via #45) with **no default type argument**. Every bare `IWalletAccount` reference — inside this module's own `.d.ts` files, in `@tetherto/wdk-wallet-evm`, and in `@tetherto/wdk-core` — now fails to compile under any consumer's TypeScript project:

```
error TS2314: Generic type 'IWalletAccount<TSignedTransaction>' requires 1 type argument(s).
```

This blocks anything that consumes those `.d.ts` files, including the WDK docs TypeScript snippets we're preparing for the policy engine documentation.

## Fix

Add a default `unknown` to the template parameter in `src/wallet-account.js`:

```diff
- * @template TSignedTransaction
+ * @template [TSignedTransaction=unknown]
```

Which emits as `IWalletAccount<TSignedTransaction = unknown>` in the `.d.ts`. Bare references (`IWalletAccount` without `<...>`) automatically resolve to `IWalletAccount<unknown>`, restoring compatibility with every existing call site.

## Scope

- 1 line of JSDoc in `src/wallet-account.js`
- 1 line in `types/src/wallet-account.d.ts` (surgically edited to avoid unrelated regeneration drift — the rest of `types/` is left as-is)

## Non-goals

The rest of `types/` shows drift when regenerated fresh under the pinned tsc 5.8.3 (interfaces emitted as classes, typedef imports rearranged, ~500 lines of noise). That's real but orthogonal to this fix — worth its own PR to run `build:types` on `main` and commit the clean baseline.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test` — 19/19 passing
- [x] Verified `.d.ts` emits `IWalletAccount<TSignedTransaction = unknown>`
- [x] Verified downstream: overlaid into `wdk-core/node_modules` and re-ran the docs snippet TS check — the "requires 1 type argument(s)" errors are gone

## Marked as draft

Opening as draft since this crosses module boundaries. Happy to iterate before it's marked ready for review.

### DIFF
--- a/src/wallet-account.js
+++ b/src/wallet-account.js
@@ -31,7 +31,7 @@ import { NotImplementedError } from './errors.js'
 
 /**
  * @interface
- * @template TSignedTransaction
+ * @template [TSignedTransaction=unknown]
  */
 export class IWalletAccount extends IWalletAccountReadOnly {
   /**
--- a/types/src/wallet-account.d.ts
+++ b/types/src/wallet-account.d.ts
@@ -1,5 +1,5 @@
 /** @interface */
-export interface IWalletAccount<TSignedTransaction> extends IWalletAccountReadOnly {
+export interface IWalletAccount<TSignedTransaction = unknown> extends IWalletAccountReadOnly {
     /**
      * The derivation path's index of this account.
      *


claudiovb (approved):

src/wallet-account.js
```diff
@@ -31,7 +31,7 @@ import { NotImplementedError } from './errors.js'
 
 /**
  * @interface
- * @template TSignedTransaction
+ * @template [TSignedTransaction=unknown]
```
Davi0kProgramsThings, line 34: The correct jsdoc for generic types with default values is:
```javascript
 * @template {unknown} TSignedTransaction
```

AlonzoRicardo: Its totally possible i mixed up the two `@template` forms, the constraint `{X} T` and the default `[T=X]` look really close, I always end up looking them up

Ran a quick comparison against tsc 5.8.3 (the version pinned in the repo) just to make sure I hadnt got it backwards:

```js
// default-syntax.js               // constraint-syntax.js
/** @template [T=unknown] */       /** @template {unknown} T */
export class Foo {}                export class Foo {}
```

Consumer, no type argument on either import:

```ts
import type { Foo as A } from './default-syntax.js'
import type { Foo as B } from './constraint-syntax.js'
declare const a: A   // OK — resolves to Foo<unknown>
declare const b: B   // TS2314: Generic type 'Foo<T>' requires 1 type argument(s).
```

Looks like the `{unknown}` form is emitting `<T extends unknown>`, so `IWalletAccount` references without a type argument still error out, which is what this PR is trying to unblock. The TS handbook uses the square bracket form under "Finally, you can specify a default for a type parameter": https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template

That said, could still be missing something on my end, was there a particular consumer setup you had in mind? is there a syntax or a case where `{unknown} T` gives you the default behavior and my test just isnt catching it?

Davi0kProgramsThings (commented):
LGTM.

PR #51, opened 2026-07-08
Add `minAmountOut` to the swidge and swap interface

nulllpc:
# Description
In cases where we want to quote first and then swidge, passing a `minAmountOut` from the quote data to the `swidge` call would ensure users are getting the quote that they approved

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

---
  - https://app.asana.com/0/0/1216404093755487

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tetherto/wdk-wallet",
-  "version": "1.0.0-beta.13",
+  "version": "1.0.0-beta.14",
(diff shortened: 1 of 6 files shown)


jonathunne (dismissed):
LGTM

jonathunne (approved):

PR #52, opened 2026-07-10
Signer: Add isDerivable and keyPair to ISigner, align IWalletAccount with signer-backed accounts

claudiovb:
# Description

This PR extends the universal signer interface and aligns the wallet account interface with it, following up on the post-merge review of tetherto/wdk-wallet-evm#40:

- **`ISigner`** gains two members:
  - `isDerivable` (`boolean`) — whether the signer supports account derivation. Lets users check derivation support without asserting on concrete types or probing whether `derive()` throws. Wallet managers already branch on this to validate default signers and to decide how to build accounts for registered signers.
  - `keyPair` (`KeyPair | null`) — the signer's key pair, or `null` for signers that cannot expose key material (e.g. hardware signers).
- **`IWalletAccount`** is updated so accounts backed by any `ISigner` can honestly fulfill the contract:
  - `keyPair` → `KeyPair | null` (null when the underlying signer doesn't allow retrieving key pairs).
  - `index` → `number | undefined` and `path` → `string | undefined` (undefined when the signer is not bound to a BIP-44 position, e.g. private-key signers).
- **`WalletManager`** constructor overloads are now annotated with `@throws`: `Error` for an invalid seed phrase on the seed overload, and `SignerError` on the signer overload when the default signer does not support account derivation (enforced by chain-specific subclasses).
- Type declarations under `types/` were hand-edited to match (no `build:types` run; the generator currently has issues — tracked separately).

No implementation code changes; runtime behavior is unchanged. Verified with the full test suite, `standard`, and a `tsc` check over the declaration files, plus end-to-end against `wdk-wallet-evm` (node and bare runtimes) with this branch as the dependency.

## Motivation and Context

Discussed in the post-merge review of tetherto/wdk-wallet-evm#40 (EVM: split signer from account). The EVM package needed `isDerivable` and `keyPair` on its `ISignerEvm` interface, and review concluded these belong on the base `ISigner` since all signer implementations should support them. Making `keyPair` nullable (and `index`/`path` optional on `IWalletAccount`) is required for signers that can never expose key material or a derivation position.

## Related Issue

No standalone issue — this implements interface changes requested in the post-merge review discussion of tetherto/wdk-wallet-evm#40.

## Type of change


eckey.pem
Davi0kProgramsThings: Quite sure this shouldn't be here. Make sure to renew your key pair and update the cybersecurity team if this can lead to sensitive content.

src/signer.js
```diff
@@ -23,7 +26,35 @@ import { NotImplementedError } from './errors.js'
...
+  }
+
+  /**
+   * The BIP 0044 derivation path.
+   *
+   * @type {string | null}
+   */
+  get path () {
+    throw new NotImplementedError('path')
+  }
```
Davi0kProgramsThings: Change this to:
```suggestion
  /**
   * The BIP 0032 derivation path.
   *
   * @type {string | null}
   */
  get path () {
    throw new NotImplementedError('path')
  }
```

src/wallet-account.js
```diff
@@ -37,33 +37,29 @@ import { NotImplementedError } from './errors.js'
...
  */
 
+/** @typedef {import('./disposable.js').IDisposable} IDisposable */
+
 /**
  * @interface
+ * @implements {IDisposable}
  * @template [TSignedTransaction=unknown]
  */
 export class IWalletAccount extends IWalletAccountReadOnly {
```
Davi0kProgramsThings: Interfaces do not implement other interfaces but rather they extend them, so here you should use @extends. In order to keep the interface more consistent, use the jsdoc to extend 'IWalletAccountReadOnly' as well:
```javascript
/**
 * @interface
 * @extends {IWalletAccountReadOnly, IDisposable}
 * @template [TSignedTransaction=unknown]
 */
export class IWalletAccount {
```

Davi0kProgramsThings: You forgot to remove the 'dispose' method from this class since now the interface inherits it from 'IDisposable'.

src/wallet-manager.js
```diff
@@ -260,14 +264,8 @@ export default class WalletManager {
...
-    this._defaultSigner?.dispose()
-
-    for (const signer of Object.values(this._signers)) {
-      signer.dispose()
-    }
-
     this._accounts = {}
-    this._defaultSigner = undefined
     this._signers = {}
+    this._defaultSigner = undefined
```
Davi0kProgramsThings: Here, the 'dispose' method should just dispose the single wallet accounts and clear the account map. No need to clear _signers and _defaultSigner as well.

tests/wallet-manager.test.js
```diff
@@ -166,18 +166,23 @@ describe('WalletManager', () => {
...
 
-      expect(disposeSpy).toHaveBeenCalledTimes(1)
+      expect(defaultDisposeSpy).not.toHaveBeenCalled()
+      expect(namedDisposeSpy).not.toHaveBeenCalled()
       expect(() => wallet.getSigner())
         .toThrow(new NoSuchElementError('No default signer set.'))
+      expect(() => wallet.getSigner('ledger'))
+        .toThrow(new NoSuchElementError('No signer found with name "ledger".'))
+      expect(wallet.getSigners()).toEqual({})
     })
```
Davi0kProgramsThings: No need to test this. In unit tests, you should just cover what the method should do rather than what the method shouldn't do. You can get rid of this test entirely or change it to assert that calling dispose on the wallet manager also disposes each of its accounts.

types/src/wallet-manager.d.ts
```diff
@@ -147,7 +150,13 @@ export default abstract class WalletManager {
...
-     * Disposes all wallet accounts and signers, clearing secret material from memory.
+     * Disposes all wallet accounts and clears references to registered signers.
+     *
+     * Signers are user-supplied and therefore not owned by the wallet manager; this
+     * method only drops the internal references so they can be garbage-collected
+     * once the caller also releases them. Downstream chain-specific wallet managers
+     * that create their own signers internally should override this method to
+     * dispose them.
      */
     dispose(): void;
```
Davi0kProgramsThings: This jsdoc is different from the documentation of the actual method in the 'WalletManager' class. You can revert this change and get rid of the 'dispose' method in this type definition though, since the additional part about signers is not really necessary and doesn't comply to KISS.

Davi0kProgramsThings (changes requested):

eckey.pem
claudiovb: Hey Davide, thanks for the catch. That was a throwaway openssl P-256 key I generated on Jul 20 to test the Cloud integration for the showcase app on a personal scratch Google project. I was working on wdk-wallet in another terminal at the same time, the key landed in the repo root and I forgot to delete it, and a git add -A during the merge commit swept it in. It was never used by anything in WDK. I'm treating it as burned anyway: removing it from the scratch project, amending the commit to drop the file, force pushing the branch and asking GitHub to purge the old commit.

types/src/wallet-manager.d.ts
claudiovb: Reverted the long jsdoc and aligned it with the one-liner on the JS method. I kept the `dispose(): void` declaration though: the abstract class is declared as `implements IDisposable` in the .d.ts, and as my understanding goes TypeScript requires a class that implements an interface to declare every member, abstract or not. Removing it would fail tsc for consumers with "Class 'WalletManager' incorrectly implements interface 'IDisposable'". Let me know if I missing something and if we should still get rid of the 'dispose' method.

eckey.pem
claudiovb: Follow-up: the file is out of the branch and the PR, and I've opened a GitHub Support ticket to drop the old unreferenced commit from the fork network so the SHA stops resolving - asking to purge the old commit. Nothing to rotate, as noted above.

types/src/wallet-manager.d.ts
Davi0kProgramsThings: This should be true for .ts files but in type definitions it should be completely possible to avoid redundant signatures when extending or implementing another type without overriding their definitions. Feel free to double check this, but it should be completely possible to get rid of the 'dispose' method here.

claudiovb: Checked this against the compiler, repro [here](https://gist.github.com/claudiovb/32fde11c36d2f58d8b85f1d1208724bf): 

The key point is that dispose() on WalletManager is not abstract. The base class implements it in JS (it disposes the accounts it owns), and the .d.ts marks getAccount, getAccountByPath and getFeeRates as abstract but not dispose, for that reason. The declaration is there to expose a method that already exists.

A .d.ts goes through the same checker as a .ts file, and implements "is only a check that the class can be treated as the interface type. It doesn't change the type of the class or its methods at all" [tsdocs](https://www.typescriptlang.org/docs/handbook/2/classes.html#cautions). So with the line removed, TypeScript users lose the method: manager.dispose() fails with TS2339, and a chain manager overriding dispose() and calling super.dispose() fails the same way. Tested on TS 5.8.3 and 5.9.3, skipLibCheck on and off. With skipLibCheck off the .d.ts itself also reports TS2720.

Only extends inherits members, which is why dropping dispose from IWalletAccount was right (its .d.ts interface extends IDisposable). We could write extends IDisposable in the WalletManager .d.ts too, but that would declare a superclass the JS class doesn't have, unless we also change wallet-manager.js to extend IDisposable at runtime. Given that, I'd keep the one-line declaration.

Davi0kProgramsThings: Yeah, my bad claudio. You are completely right, abstract classes that implement interfaces must always declare all methods whether they are abstract or not. When i left that comment i probably thought i was still reviewing the wdk-wallet-evm PR and that i was seeing 'WalletManagerEvm' which is not an abstract class. Thanks for double checking!

Davi0kProgramsThings (dismissed):
LGTM.

AlonzoRicardo (commented):
Second pass on this one, focused on the contract surface and on how it reaches the packages that implement it (4337, 7702, Safe, Squads). Verified against the branch head (`3324363`).

**Before merge**

- `IWalletAccountMultisig` (`src/multisig/wallet-account-multisig.js:53-73`, same in the `.d.ts`) still requires `index: number`, `path: string`, `keyPair: KeyPair` and doesn't extend `IDisposable`. Safe and Squads implement this interface, not `IWalletAccount`, and Safe's `index` delegates to `WalletAccountEvm.index`, which #95 removes. We're settling in Slack whether this gets folded in here or goes to a follow-up; I'll update this review once that's decided. Either way, the "Changes" section should stop saying the `index` removal is safe for every `wdk-wallet-*` package, since it isn't for the two multisig ones.

**Question**

- @claudiovb is the wdk-wallet TODO you mentioned on tetherto/wdk-wallet-evm#95 (`src/wallet-manager-evm.js:77` thread) the derivable check in the base constructor? If it becomes a follow-up issue, the `@throws {InvalidSignerError}` on the base signer overload should be dropped or enforced in the same change, since today the base documents an error it never throws.

**Checked, no issues**

- The `.d.ts` files match the JS: `IDisposable` exported from the root, `ISigner extends IDisposable`, `IWalletAccount extends IWalletAccountReadOnly, IDisposable`, `WalletManager<TSigner extends ISigner = ISigner>`, `addSigner()` returning `this`.
- `@extends {IDisposable}` on `IWalletAccount` is JSDoc-only (JS single inheritance); the `.d.ts` carries the real multi-extends. Right approach.
- No tests reference `index` or the multisig interface, so aligning `IWalletAccountMultisig` here would need no test changes.
- 4337 and 7702 are not affected by this PR on its own: their `index` becomes a chain-specific member and their managers are seed-only, so the `TSigner` generic and the `dispose()` change don't reach them. What does reach them is on #95 (the `WalletAccountEvm.index` removal and 4337's signing path); I'll comment there.

src/signer.js
```diff
@@ -23,7 +26,35 @@ import { NotImplementedError } from './errors.js'
...
+   * key material (e.g. hardware signers).
+   *
+   * @type {KeyPair | null}
+   */
+  get keyPair () {
+    throw new NotImplementedError('keyPair')
+  }
+
+  /**
+   * The BIP 0032 derivation path.
```
AlonzoRicardo: Nit: the only nullable member in the PR whose doc doesn't say when it's null (`keyPair` above and `IWalletAccount.path` both do).

```suggestion
   * The BIP 0032 derivation path, or null if the signer is not bound to a derivation position
   * (e.g. private-key signers).
```

claudiovb:
> Second pass on this one, focused on the contract surface and on how it reaches the packages that implement it (4337, 7702, Safe, Squads). Verified against the branch head (`3324363`).
> 
> **Before merge**
> 
> * `IWalletAccountMultisig` (`src/multisig/wallet-account-multisig.js:53-73`, same in the `.d.ts`) still requires `index: number`, `path: string`, `keyPair: KeyPair` and doesn't extend `IDisposable`. Safe and Squads implement this interface, not `IWalletAccount`, and Safe's `index` delegates to `WalletAccountEvm.index`, which #95 removes. We're settling in Slack whether this gets folded in here or goes to a follow-up; I'll update this review once that's decided. Either way, the "Changes" section should stop saying the `index` removal is safe for every `wdk-wallet-*` package, since it isn't for the two multisig ones.
> 
> **Question**
> 
> * @claudiovb is the wdk-wallet TODO you mentioned on [tetherto/wdk-wallet-evm#95](https://github.com/tetherto/wdk-wallet-evm/pull/95) (`src/wallet-manager-evm.js:77` thread) the derivable check in the base constructor? If it becomes a follow-up issue, the `@throws {InvalidSignerError}` on the base signer overload should be dropped or enforced in the same change, since today the base documents an error it never throws.
> 
> **Checked, no issues**
> 
> * The `.d.ts` files match the JS: `IDisposable` exported from the root, `ISigner extends IDisposable`, `IWalletAccount extends IWalletAccountReadOnly, IDisposable`, `WalletManager<TSigner extends ISigner = ISigner>`, `addSigner()` returning `this`.
> * `@extends {IDisposable}` on `IWalletAccount` is JSDoc-only (JS single inheritance); the `.d.ts` carries the real multi-extends. Right approach.
> * No tests reference `index` or the multisig interface, so aligning `IWalletAccountMultisig` here would need no test changes.
> * 4337 and 7702 are not affected by this PR on its own: their `index` becomes a chain-specific member and their managers are seed-only, so the `TSigner` generic and the `dispose()` change don't reach them. What does reach them is on #95 (the `WalletAccountEvm.index` removal and 4337's signing path); I'll comment there.

[comment link](https://github.com/tetherto/wdk-wallet/pull/52#pullrequestreview-5304134226)

Yes, that TODO is the derivable check in the base constructor. Today only WalletManagerEvm enforces it, so I dropped the @throws {InvalidSignerError} from the base signer overload in this PR and will restore it in a follow-up that adds the runtime check to WalletManager itself. The follow-up can also try a type-level bonus: typing the constructor's signer as TSigner & { isDerivable: true } so concrete signers with a literal isDerivable are rejected at compile time. It can't replace the runtime check, since interface-typed values still pass, but it catches the obvious cases early. The TSigner narrowing for addSigner and getSigner is already in this PR. I'll open the issue once we agree on this in Slack, cc @Davi0kProgramsThings

AlonzoRicardo (dismissed):

PR #54, opened 2026-07-22
Improve errors.

Davi0kProgramsThings:
# Description

PR adds new errors to all wdk-wallet classes.

## Motivation and Context

All wallet and protocol modules should share and implement the same error classes in order to make writing code on the common interfaces easier and faster. Standard errors also help client code in dispatching to the proper handler without any weird gimmick e.g., dispatching on the error message.

## Related Issue

PR fixes the following issue: -

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

---
  - https://app.asana.com/0/0/1215107647810300


src/protocols/swidge-protocol.js
```diff
@@ -263,83 +292,147 @@ export default class SwidgeProtocol {
...
+        toTokenAmount: options.tokenOutAmount,
+        minAmountOut: options.minAmountOut
+      })
+
+      const fee = result.fees.reduce((acc, f) => acc + f.amount, 0n)
+
+      return { hash: result.id, fee, tokenInAmount: result.fromTokenAmount, tokenOutAmount: result.toTokenAmount }
+    } catch (error) {
+      if (error instanceof SwidgeError) {
+        throw new SwapError(error.message, { reason: error.reason, cause: error })
```
AlonzoRicardo, line 319: When a `SwidgeError` is re-thrown as a `SwapError` here, its `reason` is copied over unchanged — but `SwidgeErrorReason` includes `SLIPPAGE_TOO_HIGH`, which isn't part of `SwapErrorReason`. The same applies in `bridge`/`quoteBridge`, where `BridgeErrorReason` is missing both `COULD_NOT_MET_THRESHOLD` and `SLIPPAGE_TOO_HIGH`. As it stands a `SwapError` or `BridgeError` can carry a reason that isn't a member of its own enum. Could we align the reason sets (adding the slippage/threshold reasons where they apply) or translate the reason when re-wrapping?

src/protocols/swap-protocol.js
```diff
@@ -134,6 +163,11 @@ export default class SwapProtocol {
    * @abstract
    * @param {SwapOptions} options - The swap's options.
    * @returns {Promise<Omit<SwapResult, 'hash'>>} The swap's quotes.
+   * @throws {ValueError} If the swap options are not valid.
```
AlonzoRicardo, line 166: `ISwapProtocol.quoteSwap` lists `ReadOnlyAccountRequiredError` among the errors it can throw, but this abstract override doesn't. Since implementations will still throw it, could we add the matching `@throws {ReadOnlyAccountRequiredError}` here so the abstract class stays in step with the interface?

src/protocols/bridge-protocol.js
```diff
@@ -117,6 +146,11 @@ export default class BridgeProtocol {
    * @abstract
    * @param {BridgeOptions} options - The bridge's options.
    * @returns {Promise<Omit<BridgeResult, 'hash'>>} The bridge's quotes.
+   * @throws {ValueError} If the bridge options are not valid.
```
AlonzoRicardo, line 149: Same point as `quoteSwap`: `IBridgeProtocol.quoteBridge` declares `ReadOnlyAccountRequiredError`, but this abstract override leaves it out. Could we add the matching `@throws {ReadOnlyAccountRequiredError}` line here?

src/protocols/lending-protocol.js
```diff
@@ -199,6 +270,11 @@ export default class LendingProtocol {
    * @abstract
    * @param {SupplyOptions} options - The supply's options.
    * @returns {Promise<Omit<SupplyResult, 'hash'>>} The supply's costs.
+   * @throws {ValueError} If the supply options are not valid.
```
AlonzoRicardo, line 273: All four quote methods here — `quoteSupply`, `quoteWithdraw`, `quoteBorrow`, `quoteRepay` — omit `@throws {ReadOnlyAccountRequiredError}`, even though `ILendingProtocol` declares it for each. Could we add the missing line to all four so the abstract class matches its interface?

src/protocols/swidge-protocol.js
Davi0kProgramsThings: 'SLIPPAGE_TOO_HIGH' can never be thrown since we are not using nor providing a value for the 'slippage' argument when calling 'swidge' inside the override of the 'swap' method. The same applies with the 'bridge' method and the COULD_NOT_MET_THRESHOLD error as well. We are not providing a value for the 'minAmountOut' option so such exception will never be thrown.

src/protocols/swap-protocol.js
Davi0kProgramsThings: The abstract class forces at least a read-only account to be set at construction. So, the quote swap method will never throw 'ReadOnlyAccountRequiredError' in implementation that extend 'SwapProtocol'. This is due to the constructors of the abstract class, which don't allow creating a swap protocol without an account:
https://github.com/Davi0kProgramsThings/wdk-wallet/blob/feature/errors/src/protocols/swap-protocol.js#L124

For protocols that allow such constructor, i kept the @throws {ReadOnlyAccountRequiredError} [...] tag in the abstract class:
https://github.com/Davi0kProgramsThings/wdk-wallet/blob/feature/errors/src/protocols/swidge-protocol.js#L338

src/protocols/bridge-protocol.js
Davi0kProgramsThings: https://github.com/tetherto/wdk-wallet/pull/54#discussion_r3639214889

src/protocols/lending-protocol.js
Davi0kProgramsThings: https://github.com/tetherto/wdk-wallet/pull/54#discussion_r3639214889

src/errors.js
```diff
@@ -13,69 +13,314 @@
...
+  /**
+   * Creates a new invalid signer error.
+   *
+   * @param {string} message - The error's message.
+   * @param {ErrorOptions} [options] - The error's options.
+   */
+  constructor (message, options) {
+    super(message, options)
+
+    this.name = 'SignerError'
```
AlonzoRicardo: Other classes set this.name equal to the class name, shouldn't this be "InvalidSignerError" instead of "SignerError"?

src/protocols/fiat-protocol.js
```diff
@@ -232,68 +284,116 @@ export default class FiatProtocol {
...
+   *
+   * @abstract
    * @param {BuyOptions} options - The options for the buy operation.
    * @returns {Promise<BuyResult>} The URL for the user to complete the purchase.
+   * @throws {AccountRequiredError} If the protocol requires a full account to perform a purchase.
+   * @throws {ValueError} If the buy options are not valid.
+   * @throws {ProviderRequiredError} If the method requires a provider.
+   * @throws {ProviderError} If the provider fails to perform the purchase.
+   * @throws {BuyError} If the purchase fails with an error.
+   * @throws {MaximumFeeExceededError} If the the costs of the purchase exceeds the buy max. fee option.
```
AlonzoRicardo, line 312: is a fee-limit config intended here?, This @throws {MaximumFeeExceededError} refers to a "buy max. fee option"?, i say that because FiatProtocol takes no config and defines no such option so im not sure

src/protocols/errors.js
```diff
@@ -0,0 +1,468 @@
...
+
+/**
+ * Thrown when a supply fails with an error.
+ */
+export class SupplyError extends WdkError {
+  /**
+   * Creates a new supply error.
+   *
+   * @param {string} message - The error's message.
+   * @param {BridgeErrorOptions & ErrorOptions} options - The error's options.
```
AlonzoRicardo: should be `SupplyErrorOptions` so `reason` is typed against the supply enum.?

src/errors.js
Davi0kProgramsThings: Yes, nice catch.

src/protocols/fiat-protocol.js
Davi0kProgramsThings: Yes, that's to allow protocol implementations to introduce a buy max. fee threshold and throw the proper error. If we don't do so, sub-classes won't be able to throw on such condition since the specification of the super-class would not allow it. This way, they can introduce this option and check without breaking SOLID. As for all errors and @throws tags, implementations are still free to never throw it if the implementation allows it.

src/protocols/errors.js
Davi0kProgramsThings: Yes, nice catch!

src/protocols/errors.js
```diff
@@ -0,0 +1,468 @@
...
+ * @property {SwidgeErrorReason} reason - The error's reason.
+ */
+
+/**
+ * Enum for swap error reasons.
+ *
+ * @readonly
+ * @enum {string}
+ */
+export const SwapErrorReason = {
```
nulllpc, line 76: The values are duplicated for the reason enums, I don't think we need this level of separation

src/errors.js
```diff
@@ -13,69 +13,314 @@
...
+ * Enum for provider error reasons.
+ *
+ * @readonly
+ * @enum {number}
+ */
+export const ProviderErrorReason = {
+  /**
+   * Thrown when the client fails to establish a connection with the provider.
+   */
+  NETWORK_ERROR: 0,
```
nulllpc: Can implementers add their own reasons? If yes, then I feel like we cannot control the numbers here, different implementers might use the same number. If no, then we would be in charge of managing the enum?

src/protocols/swidge-protocol.js
```diff
@@ -263,83 +292,147 @@ export default class SwidgeProtocol {
...
+        fromTokenAmount: options.tokenInAmount,
+        toTokenAmount: options.tokenOutAmount,
+        minAmountOut: options.minAmountOut
+      })
+
+      const fee = result.fees.reduce((acc, f) => acc + f.amount, 0n)
+
+      return { fee, tokenInAmount: result.fromTokenAmount, tokenOutAmount: result.toTokenAmount }
+    } catch (error) {
+      if (error instanceof SwidgeError) {
```
nulllpc, line 353: As discussed, I don't think we can rely on `instanceof`. The biggest risk is that different modules might have different versions of `wdk-wallet`, so it would fail silently. It also links to my other comment about having only 1 class `WdkError`, then we can differentiate using fields instead

src/protocols/errors.js
Davi0kProgramsThings: I did this since i am expecting those enums to grow each with different reasons before merging the PR. Separating each reason also helps with providing error-specific descriptions. Also, since modules can extend the available reasons it makes sense to allow them to import a different enum for each error.

src/errors.js
Davi0kProgramsThings: Yes, they can. The design is meant to allow modules to add reasons without breaking SOLID. They can do so by extending the enum and the corresponding *Error class. You have a point here, nice catch. I will change 'ProviderErrorReason' to map strings just like all the other enums:
```suggestion
  NETWORK_ERROR: 'NETWORK_ERROR',
```

src/protocols/swidge-protocol.js
Davi0kProgramsThings: This is quite an edge case, since ideally the last version of each module should point to the same release of wdk-wallet. This is something we can take care of ourself, so we shouldn't really design the errors around it. Also, users can still rely on the error's name to dispatch to the proper handler (which is actually very common):
```suggestion
      if (error.name === 'SwidgeError') {
```

src/errors.js
```diff
@@ -13,77 +13,324 @@
...
+ * Enum for provider error reasons.
+ *
+ * @readonly
+ * @enum {number}
+ */
+export const ProviderErrorReason = {
+  /**
+   * Thrown when the client fails to establish a connection with the provider.
+   */
+  NETWORK_ERROR: 'NETWORK_ERROR',
```
AlonzoRicardo, line 41: These reasons are now string values (`'NETWORK_ERROR'`, etc.), but the enum is still tagged `@enum {number}` just above (line 35), and `ProviderError.reason` is still `@type {number}` (line 257). Could we switch both to `{string}` so they match the values and the other reason-bearing errors?

Davi0kProgramsThings: Nice catch!

jonathunne (approved):

PR #56, opened 2026-08-04
Add waitForTransaction method

Boka44:
# Description
Adds a shared `waitForTransaction` helper to the base wallet account. It polls
`getTransaction` until the transaction reaches the requested finality target
(`confirmed` or `final`), and surfaces clear errors when a transaction reverts,
is dropped, or times out.

To avoid a single RPC blip aborting a long wait, polling now tolerates a few
consecutive `getTransaction` failures before giving up. This is configurable via
a new `maxPollErrors` option (default: 3).

## Motivation and Context
Each chain module previously had to implement its own logic for waiting on a
transaction. Centralizing it here gives every module consistent finality
handling and error reporting, and makes waits resilient to transient node
hiccups.

## Type of change

---
  - https://app.asana.com/0/0/1217123489895949

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -82,19 +82,14 @@ export class NoSuchElementError extends Error {
 
 export class TimeoutError extends Error {
   /**
-   * Create a new timeout error. Thrown when a transaction does not reach the
-   * requested finality target within the timeout.
+   * Create a new timeout error. Thrown when an operation does not complete
+   * within its allotted time.
    *
-   * @param {string} hash - The transaction's hash.
-   * @param {string} target - The requested finality target.
-   * @param {import('./wallet-account-read-only.js').TransactionReceipt | null} [receipt] - The last-seen receipt, or null if never seen.
+   * @param {string} message - The error's message.
    */
-  constructor (hash, target, receipt = null) {
-    super(`Transaction '${hash}' did not reach '${target}' within the timeout.`)
+  constructor (message) {
+    super(message)
 
     this.name = 'TimeoutError'
-
-    /** @type {import('./wallet-account-read-only.js').TransactionReceipt | null} */
-    this.receipt = receipt
   }
 }
--- a/src/errors.js
+++ b/src/errors.js
@@ -80,46 +80,10 @@ export class NoSuchElementError extends Error {
   }
 }
 
-export class TransactionFailedError extends Error {
+export class TimeoutError extends Error {
   /**
-   * Create a new transaction failed error. Thrown when a transaction lands on
-   * chain but its execution reverts.
-   *
-   * @param {string} hash - The transaction's hash.
-   * @param {import('./wallet-account-read-only.js').TransactionReceipt} [receipt] - The failed transaction's receipt.
-   */
-  constructor (hash, receipt) {
-    super(`Transaction '${hash}' failed.`)
-
-    this.name = 'TransactionFailedError'
-
-    /** @type {import('./wallet-account-read-only.js').TransactionReceipt | undefined} */
-    this.receipt = receipt
-  }
-}
-
-export class TransactionDroppedError extends Error {
-  /**
-   * Create a new transaction dropped error. Thrown when a transaction is evicted
-   * or replaced and never lands on chain.
-   *
-   * @param {string} hash - The transaction's hash.
-   * @param {import('./wallet-account-read-only.js').TransactionReceipt} [receipt] - The dropped transaction's receipt.
-   */
-  constructor (hash, receipt) {
-    super(`Transaction '${hash}' was dropped.`)
-
-    this.name = 'TransactionDroppedError'
-
-    /** @type {import('./wallet-account-read-only.js').TransactionReceipt | undefined} */
-    this.receipt = receipt
-  }
-}
-
-export class TransactionConfirmationTimeoutError extends Error {
-  /**
-   * Create a new transaction confirmation timeout error. Thrown when a
-   * transaction does not reach the requested finality target within the timeout.
+   * Create a new timeout error. Thrown when a transaction does not reach the
+   * requested finality target within the timeout.
    *
    * @param {string} hash - The transaction's hash.
    * @param {string} target - The requested finality target.
@@ -128,7 +92,7 @@ export class TransactionConfirmationTimeoutError extends Error {
   constructor (hash, target, receipt = null) {
     super(`Transaction '${hash}' did not reach '${target}' within the timeout.`)
 
-    this.name = 'TransactionConfirmationTimeoutError'
+    this.name = 'TimeoutError'
 
     /** @type {import('./wallet-account-read-only.js').TransactionReceipt | null} */
     this.receipt = receipt
--- a/src/errors.js
+++ b/src/errors.js
@@ -79,3 +79,58 @@ export class NoSuchElementError extends Error {
     this.name = 'NoSuchElementError'
   }
 }
+
+export class TransactionFailedError extends Error {
+  /**
+   * Create a new transaction failed error. Thrown when a transaction lands on
+   * chain but its execution reverts.
+   *
+   * @param {string} hash - The transaction's hash.
+   * @param {import('./wallet-account-read-only.js').TransactionReceipt} [receipt] - The failed transaction's receipt.
+   */
+  constructor (hash, receipt) {
+    super(`Transaction '${hash}' failed.`)
+
+    this.name = 'TransactionFailedError'
+
+    /** @type {import('./wallet-account-read-only.js').TransactionReceipt | undefined} */
+    this.receipt = receipt
+  }
+}
+
+export class TransactionDroppedError extends Error {
+  /**
+   * Create a new transaction dropped error. Thrown when a transaction is evicted
+   * or replaced and never lands on chain.
+   *
+   * @param {string} hash - The transaction's hash.
+   * @param {import('./wallet-account-read-only.js').TransactionReceipt} [receipt] - The dropped transaction's receipt.
+   */
+  constructor (hash, receipt) {
+    super(`Transaction '${hash}' was dropped.`)
+
+    this.name = 'TransactionDroppedError'
+
+    /** @type {import('./wallet-account-read-only.js').TransactionReceipt | undefined} */
+    this.receipt = receipt
+  }
+}
+
+export class TransactionConfirmationTimeoutError extends Error {
+  /**
+   * Create a new transaction confirmation timeout error. Thrown when a
+   * transaction does not reach the requested finality target within the timeout.
+   *
+   * @param {string} hash - The transaction's hash.
+   * @param {string} target - The requested finality target.
+   * @param {import('./wallet-account-read-only.js').TransactionReceipt | null} [receipt] - The last-seen receipt, or null if never seen.
+   */
+  constructor (hash, target, receipt = null) {
+    super(`Transaction '${hash}' did not reach '${target}' within the timeout.`)
+
+    this.name = 'TransactionConfirmationTimeoutError'
+
+    /** @type {import('./wallet-account-read-only.js').TransactionReceipt | null} */
+    this.receipt = receipt
+  }
+}
--- a/src/wallet-account-read-only.js
+++ b/src/wallet-account-read-only.js
@@ -93,8 +93,8 @@ export const FINALITY = {
 /**
  * @typedef {Object} WaitForTransactionOptions
  * @property {WaitForTransactionTarget} [target] - The finality target to wait for (default: 'confirmed').
- * @property {number} [timeout] - The total time budget in milliseconds before giving up (default: per-module).
- * @property {number} [interval] - The poll cadence in milliseconds (default: per-module).
+ * @property {number} [timeout] - The total time budget in milliseconds before giving up (default: 60000).
+ * @property {number} [interval] - The poll cadence in milliseconds (default: 4000).
  * @property {number} [maxPollErrors] - How many consecutive getTransaction() failures to tolerate before rethrowing (default: 3).
  */
 
@@ -161,6 +161,24 @@ export class IWalletAccountReadOnly extends IWalletAccountReadOnlySimple {
  * @implements {IWalletAccountReadOnly}
  */
 export default class WalletAccountReadOnly {
+  /**
+   * The default poll cadence for {@link waitForTransaction}, in milliseconds.
+   * Subclasses may override this to match their finality expectations.
+   *
+   * @protected
+   * @type {number}
+   */
+  static _DEFAULT_WAIT_INTERVAL = 4000
+
+  /**
+   * The default total time budget for {@link waitForTransaction}, in milliseconds.
+   * Subclasses may override this to match their finality expectations.
+   *
+   * @protected
+   * @type {number}
+   */
+  static _DEFAULT_WAIT_TIMEOUT = 60000
+
   /**
    * Creates a new read-only wallet account.
    *
@@ -292,8 +310,8 @@ export default class WalletAccountReadOnly {
   async waitForTransaction (hash, options = {}) {
     const {
       target = 'confirmed',
-      interval = this._defaultWaitInterval,
-      timeout = this._defaultWaitTimeout,
+      interval = this.constructor._DEFAULT_WAIT_INTERVAL,
+      timeout = this.constructor._DEFAULT_WAIT_TIMEOUT,
       maxPollErrors = 3
     } = options
 
@@ -337,26 +355,5 @@ export default class WalletAccountReadOnly {
 
       await new Promise(resolve => setTimeout(resolve, interval))
     }
-  }
-
-  /**
-  * The default poll cadence for {@link waitForTransaction}, in milliseconds.
-  *
-  * @protected
-  * @type {number}
-  */
-  get _defaultWaitInterval () {
-    return 4000
-  }
-
-  /**
-   * The default total time budget for {@link waitForTransaction}, in milliseconds.
-   * Modules override this to match their finality expectations.
-   *
-   * @protected
-   * @type {number}
-   */
-  get _defaultWaitTimeout () {
-    return 60000
   }
 }
--- a/src/wallet-account-read-only.js
+++ b/src/wallet-account-read-only.js
@@ -62,7 +62,7 @@ import { IWalletAccountReadOnlySimple } from './wallet-account-read-only-simple.
  * transaction and receipt objects, etc.).
  *
  * @typedef {Object} TransactionReceipt
- * @property {string} id - The transaction's identifier (hash / signature / lt:hash).
+ * @property {string} hash - The transaction's identifier (hash / signature / lt:hash).
  * @property {Finality} finality - The transaction's finality level.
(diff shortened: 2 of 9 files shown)


src/wallet-account-read-only.js
```diff
@@ -40,6 +45,44 @@ import { NotImplementedError } from './errors.js'
...
+
+/**
+ * A normalized, cross-chain transaction receipt. Blockchain modules extend this
+ * type with their own native receipt fields (e.g. `confirmations`, the raw
+ * transaction and receipt objects, etc.).
+ *
+ * @typedef {Object} TransactionReceipt
+ * @property {string} id - The transaction's identifier (hash / signature / lt:hash).
+ * @property {Finality} finality - The transaction's finality level.
+ * @property {boolean | null} success - The execution result, or null while pending/dropped.
```
Davi0kProgramsThings: It would be better for 'success' to be undefined when the transaction is still pending/dropped (instead of null):
```suggestion
 * @property {boolean} [success] - The execution's result (not set if the transaction is still pending or it has been dropped).
```

src/wallet-account-read-only.js
```diff
@@ -40,6 +45,44 @@ import { NotImplementedError } from './errors.js'
...
+ * @property {string | number} [blockRef] - A reference to the including block (block hash / slot / masterchain seqno).
```
Davi0kProgramsThings: It would better to consistently return a single type for the block ref field. If we can use number on all implementations, then we should prefer it over string:
```suggestion
 * @property {number} [blockRef] - A reference to the including block (block hash / slot / masterchain seqno).
```

Davi0kProgramsThings: What about using just 'block'?
```suggestion
 * @property {number} [block] - A reference to the including block (block hash / slot / masterchain seqno).
```

src/wallet-account-read-only.js
```diff
@@ -105,12 +148,34 @@ export class IWalletAccountReadOnly {
...
+
+  /**
+   * Returns a normalized, finality-based receipt for a transaction.
+   *
+   * @param {string} hash - The transaction's identifier (hash / signature / lt:hash).
+   * @returns {Promise<TransactionReceipt | null>} The normalized receipt, or null if the transaction is not known.
+   */
+  async getTransaction (hash) {
+    throw new NotImplementedError('getTransaction(hash)')
+  }
```
Davi0kProgramsThings: It might be worth renaming the 'hash' argument to 'id', since the 'TransactionReceipt' type refers to it with the latter:
```suggestion
  /**
   * Returns a normalized, finality-based receipt for a transaction.
   *
   * @param {string} id - The transaction's identifier (hash / signature / lt:hash).
   * @returns {Promise<TransactionReceipt | null>} The normalized receipt, or null if the transaction is not known.
   */
  async getTransaction (id) {
    throw new NotImplementedError('getTransaction(id)')
  }
```

Davi0kProgramsThings: We were discussing about making this method throw if no transaction receipt has been found. We can make it throw a 'NoSuchElementError' and change its specification to:
```suggestion
  /**
   * Returns a normalized, finality-based receipt for a transaction.
   *
   * @param {string} id - The transaction's identifier (hash / signature / lt:hash).
   * @returns {Promise<TransactionReceipt>} The normalized receipt.
   * @throws {ValueError} If the id is not a valid identifier.
   * @throws {NoSuchElementError} If no transaction has been found for the given id.
   */
  async getTransaction (id) {
    throw new NotImplementedError('getTransaction(id)')
  }
```

src/wallet-account-read-only.js
```diff
@@ -40,6 +45,44 @@ import { NotImplementedError } from './errors.js'
...
+/**
+ * The finality level to wait for.
+ *
+ * @typedef {'confirmed' | 'final'} WaitForTransactionTarget
+ */
+
+/**
+ * @typedef {Object} WaitForTransactionOptions
+ * @property {WaitForTransactionTarget} [target] - The finality target to wait for (default: 'confirmed').
+ * @property {number} [timeout] - The total time budget in milliseconds before giving up (default: per-module).
```
Davi0kProgramsThings: It's better to choose a default value up-front instead of letting each module define it. Use whatever value sounds more reasonable to you. The same applies for the 'interval' field as well.

src/wallet-account-read-only.js
```diff
@@ -210,11 +275,128 @@ export default class WalletAccountReadOnly {
...
+   * interprets the normalized receipt returned by {@link getTransaction}.
+   *
+   * @param {string} hash - The transaction's identifier.
+   * @param {WaitForTransactionOptions} [options] - The wait options.
+   * @returns {Promise<TransactionReceipt>} The terminal receipt once the target is reached.
+   * @throws {TransactionFailedError} If the transaction lands but reverts (success === false). The receipt is exposed on `.receipt`.
+   * @throws {TransactionDroppedError} If the transaction is reported dropped on two consecutive polls. The receipt is exposed on `.receipt`.
+   * @throws {TransactionConfirmationTimeoutError} If the target is not reached before the timeout. The last-seen receipt (or null) is exposed on `.receipt`.
+   */
+  async waitForTransaction (hash, options = {}) {
```
Davi0kProgramsThings, line 255: The only reason this method should throw an error is timeouts. Making it throw on failed or dropped transactions is bad design. Instead, the method should just return the receipt whenever its status changes to 'dropped', or to the given options.target. The user can then dispatch on the receipt's status instead of catching the proper exception. The same also applies to transactions that reverts: the user can just check and dispatch on the success field. That's the exact reason these two fields exists at all.
```suggestion
  /**
   * Blocks until a transaction reaches the requested finality target, fails, is dropped, or times out.
   *
   * The polling loop and target resolution are chain-agnostic: this method only
   * interprets the normalized receipt returned by {@link getTransaction}.
   *
   * @param {string} id - The transaction's identifier.
   * @param {WaitForTransactionOptions} [options] - The wait options.
   * @returns {Promise<TransactionReceipt>} The terminal receipt once the target is reached.
   * @throws {TimeoutError} If the transaction is still pending and the operation times out.
   */
  async waitForTransaction (id, options = {}) {
```

src/wallet-account-read-only.js
```diff
@@ -105,12 +148,34 @@ export class IWalletAccountReadOnly {
...
+  /**
+   * Blocks until a transaction reaches the requested finality target, fails, is dropped, or times out.
+   *
+   * @param {string} hash - The transaction's identifier.
+   * @param {WaitForTransactionOptions} [options] - The wait options.
+   * @returns {Promise<TransactionReceipt>} The terminal receipt.
+   */
+  async waitForTransaction (hash, options) {
+    throw new NotImplementedError('waitForTransaction(hash, options)')
+  }
```
Davi0kProgramsThings: It might be worth renaming the 'hash' argument to 'id', since the 'TransactionReceipt' type refers to it with the latter:
```suggestion
  /**
   * Blocks until a transaction reaches the requested finality target, fails, is dropped, or times out.
   *
   * @param {string} id - The transaction's identifier.
   * @param {WaitForTransactionOptions} [options] - The wait options.
   * @returns {Promise<TransactionReceipt>} The terminal receipt.
   */
  async waitForTransaction (id, options) {
    throw new NotImplementedError('waitForTransaction(id, options)')
  }
```

src/errors.js
```diff
@@ -79,3 +79,58 @@ export class NoSuchElementError extends Error {
...
+   */
+  constructor (hash, target, receipt = null) {
+    super(`Transaction '${hash}' did not reach '${target}' within the timeout.`)
+
+    this.name = 'TransactionConfirmationTimeoutError'
+
+    /** @type {import('./wallet-account-read-only.js').TransactionReceipt | null} */
+    this.receipt = receipt
+  }
+}
```
Davi0kProgramsThings: This error can just be:
```suggestion
/**
 * Thrown when an operation times out.
 */
export class TimeoutError extends Error {
  /**
   * Create a new timeout error.
   *
   * @param {string} message - The error's message.
   */
  constructor (message) {
    super(message)

    this.name = 'TimeoutError'
  }
}
```

src/wallet-account-read-only.js
```diff
@@ -105,12 +148,34 @@ export class IWalletAccountReadOnly {
...
+
+  /**
+   * Blocks until a transaction reaches the requested finality target, fails, is dropped, or times out.
+   *
+   * @param {string} hash - The transaction's identifier.
+   * @param {WaitForTransactionOptions} [options] - The wait options.
+   * @returns {Promise<TransactionReceipt>} The terminal receipt.
+   */
+  async waitForTransaction (hash, options) {
```
Davi0kProgramsThings: Sub-types cannot throw more errors than their super-type allows. For this reason, you should mark that this method can throw 'TimeoutError' on the interface as well:
```suggestion
  /**
   * Blocks until a transaction reaches the requested finality target, fails, is dropped, or times out.
   *
   * @param {string} hash - The transaction's identifier.
   * @param {WaitForTransactionOptions} [options] - The wait options.
   * @returns {Promise<TransactionReceipt>} The terminal receipt.
   * @throws {TimeoutError} If the transaction is still pending and the operation times out.
   */
  async waitForTransaction (hash, options) {
```

src/wallet-account-read-only.js
```diff
@@ -210,11 +275,128 @@ export default class WalletAccountReadOnly {
...
+  /**
+   * The default poll cadence for {@link waitForTransaction}, in milliseconds.
+   * Modules override this to match their block cadence.
+   *
+   * @protected
+   * @type {number}
+   */
+  get _defaultWaitInterval () {
+    return 4000
+  }
```
Davi0kProgramsThings: There's no need to state that modules can/should override this to match their block cadence:
 ```suggestion
  /**
   * The default poll cadence for {@link waitForTransaction}, in milliseconds.
   *
   * @protected
   * @type {number}
   */
  get _defaultWaitInterval () {
    return 4000
  }
```

File: src/wallet-account-read-only.js [hunk 10]
```diff
@@ -210,11 +275,128 @@ export default class WalletAccountReadOnly {
...
+   * @param {WaitForTransactionTarget} target - The requested finality target.
+   * @returns {boolean} True if the observed finality satisfies the target.
+   */
+  _meetsFinality (finality, target) {
+    if (target === 'final') {
+      return finality === 'final'
+    }
+
+    return finality === 'confirmed' || finality === 'final'
+  }
```
Davi0kProgramsThings: You can replace this method by defining a static member that maps each finality to a number.
```javascript
export default class WalletAccountReadOnly {
  /**
   * A map that assigns a unique number to each available finality.
   *
   * @protected
   * @type {Record<Finality, number>}
   */
  static _FINALITY = {
    'pending': 0,
    'dropped': 1,
    'confirmed': 2,
    'final': 3
  }
```

This way, you can then check if the transaction's status meets the target with the following and drop the '_meetsFinality' method:
```javascript
if (_FINALITY[receipt.finality] >= _FINALITY[target]) {
  [...]
}
```

tests/wallet-account-read-only.test.js
```diff
@@ -41,4 +73,104 @@ describe('WalletAccountReadOnly', () => {
         .rejects.toThrow("The account's address must be set to perform this operation.")
     })
   })
+
+  describe('getTransaction', () => {
+    test('should throw NotImplementedError by default', async () => {
+      const account = new DummyWalletAccountReadOnly(ADDRESS)
+      await expect(account.getTransaction(HASH)).rejects.toThrow("Method 'getTransaction(hash)' must be implemented.")
+    })
+  })
```
Davi0kProgramsThings: No need to cover this test case.

Davi0kProgramsThings (changes requested):

src/wallet-account-read-only.js
Boka44: Everything returns a number except for EVM, which returns a string. I would leave as is

src/wallet-account-read-only.js
Boka44: We should keep hash as the param, as it is a convention we use across the entire wdk. `getTransactionReceipt` also uses hash, as a example.

Boka44: I had my clanker format a proper response:

Two separate things here though:

Invalid hash → ValueError: totally agree, easy win. I'll standardize and document that.

Not-found → NoSuchElementError instead of null: this one's trickier, because the null return isn't just a convenience — it's what waitForTransaction uses to mean "not seen yet, keep polling." If getTransaction throws instead, a freshly-broadcast tx (not indexed yet) would throw on the first few polls and the wait would give up early. So we can't just flip it without also teaching the poll loop to treat "not found" as the keep-polling case.

There's also a philosophical angle: getTransaction is our polling primitive, and "not found yet" is a normal, expected state rather than a real error — so throwing on every poll feels a bit like using exceptions for control flow.

A few ways we could go:

Option A — keep null for not-found (my lean). It reads naturally for a lookup and plays nicely with polling. We still add ValueError for bad input.
Option B — throw NoSuchElementError, full sweep. Refactor waitForTransaction to treat it as "keep polling," and apply it across all 12 modules + types + tests. Consistent with the SDA lookups, just more moving parts.

src/wallet-account-read-only.js
Boka44: Well, the intervals and timeouts do vary per chain. While we could  find a somewhat reasonable number, we have a huge outlier in BTC. That is why I opted per chain, as a default value that satisfies BTC would be absolute overkill on any other chain.

src/wallet-account-read-only.js
Boka44: ok good call on this one

src/wallet-account-read-only.js
Boka44: It's not a bad idea, though I am not particular on establishing an order where dropped is a step above pending.. technically the only true targets are confirmed and final anyways. I think it is cleaner as is.

Davi0kProgramsThings: Both options work fine. Consider the following though: _FINALITY is not meant to order statuses, it just enumerates them. In fact, instead of defining a static member we could make it a global enum and also export it out of the module so that clients can use it as well. External code wouldn't be able to change it, so it would be safe to use the above check. It's a quick trick to avoid an extra method and assist the implementation of 'waitForTransaction'. Cleanliness and readability should be fine either ways:
```javascript
/**
 * Enum for available finalities.
 *
 * @readonly
 * @enum {number}
 */
export const FINALITY = {
  'pending': 0,
  'dropped': 1,
  'confirmed': 2,
  'final': 3
}

export default class WalletAccountReadOnly {
```

src/wallet-account-read-only.js
Davi0kProgramsThings: Can't we just use 60,000 and 4,000 like 'WalletAccountReadOnly'? The issue with not setting a default value for those fields is that code that has been written against the interface will not know how much they will need to wait when they are not explicitly provide them. In fact, sub-classes shouldn't change the default value of an argument so ideally we should also drop the two getter methods and replace them with protected static members just like we are doing [here](https://github.com/tetherto/wdk-wallet-evm/blob/main/src/wallet-manager-evm.js#L42).

If this is totally unreasonable, then we can ignore this flaw. In such case, i would drop the "... (default: per-module)" part of the descriptions since they are quite unusual. Not providing a default value will be enough to state that each sub-class define its own. Let's try to find some default values that can work for all modules though, maybe the sweet spot is somewhere near 60,000 and 4,000.

src/wallet-account-read-only.js
Davi0kProgramsThings: Don't ethereum and ethereum layer-2 blockchains use numeric block identifiers as well? We can just convert the string into a number. I guess you are referring to some special values like 'latest'. In such case, we shouldn't really use them and instead we should convert them to the corresponding block number. 'latest' loses its meaning very quickly and doesn't allow users to store receipts in a persistent storage such as a DB.

src/wallet-account-read-only.js
Davi0kProgramsThings: This makes sense, but in such case we should replace 'id' with 'hash' in the 'TransactionReceipt' type as well then. I'm good with either options. 'getTransactionReceipt' has been deprecated so we are free to change the argument name in the method that will replace it.

Davi0kProgramsThings: I would go with option B. This will make the behavior of the method on par with the other interfaces and the new error handling. You just need to change the implementation of 'waitForTransaction' in order to try-catch the calls to 'getTransaction' and retry if a 'NoSuchElementError' has been thrown (basically behaving like the method returned null).

src/wallet-account-read-only.js
Boka44: Yeah you are right. EVM was using blockhash instead of block number. All good now

src/wallet-account-read-only.js
Boka44: OK I found a happy medium. Set up defaults that please most modules (minus btc), and allow a few modules to override the defaults with their own.

src/wallet-account-read-only.js
Davi0kProgramsThings, line 1: All these changes should have been made to wallet-account-read-only-simple.js and not wallet-account-read-only.js. You just need to revert all changes to this file except for the 'WalletAccountReadOnly' abstract class and move them to the other file. As you can see, you are duplicating the 'getTransactionReceipt' method since it already exists in the 'IWalletAccountReadOnlySimple' interface so that's the one we should actually deprecate. Remember to update the .d.ts files as well.

src/wallet-account-read-only.js
```diff
@@ -162,11 +271,89 @@ export default class WalletAccountReadOnly {
...
+        }
+      } else {
+        droppedStreak = 0
+      }
+
+      if (Date.now() >= deadline) {
+        throw new TimeoutError(`Transaction '${hash}' did not reach '${target}' within the timeout.`)
+      }
+
+      await new Promise(resolve => setTimeout(resolve, interval))
```
Davi0kProgramsThings: Just a minor suggestion: define a sleep utility function at the top-level of the file to use in this method.
```javascript
async function sleep(amount) {
  await new Promise(resolve => setTimeout(resolve, amount))
}

export default class WalletAccountReadOnly {
  async waitForTransaction (hash, options = {}) {
    [...]

    while (true) {
      [...]

      await sleep(interval)
    }
  }


```

Davi0kProgramsThings (changes requested):

src/wallet-account-read-only.js
Boka44: @Davi0kProgramsThings do you mean switch everything over, including the waitForTransaction logic? Also these classes are defined as interfaces, but since one of them will have logic I am not sure about keeping it defined as an interface

Davi0kProgramsThings: The 'WalletAccountReadOnly' abstract class should remain the same. You just need to move the new type definitions and methods from 'IWalletAccountReadOnly' to 'IWalletAccountReadOnlySimple'. The default implementation of the wait for transaction method should stay where it already is.

Davi0kProgramsThings (dismissed):
LGTM.

Davi0kProgramsThings (changes requested):
LGTM.

Davi0kProgramsThings (approved):
LGTM.

PR #60, opened 2026-08-06
Add auto execute's transaction result to return types.

Davi0kProgramsThings:
# Description

PR adds auto execute's transaction result to the return type of the 'propose' and 'approveProposal' methods in the 'IWalletAccountMultisig' interface.

## Motivation and Context

Without this PR, users that set auto execute to true had no way to fetch the corresponding transaction's hash and fee. The pull request adds an optional 'transaction' field to the return type of the 'propose' and 'approveProposal' methods that holds such data. This field is set only if auto execute has been enabled and the method call triggers the execution of the tx.

## Related Issue

PR fixes the following issue: -

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

---
  - https://app.asana.com/0/0/1217218764605753

### DIFF
--- a/src/multisig/index.js
+++ b/src/multisig/index.js
@@ -19,6 +19,7 @@
 /** @typedef {import('../wallet-account.js').KeyPair} KeyPair */
 
 /** @typedef {import('./wallet-account-multisig.js').MultisigTransactionOptions} MultisigTransactionOptions */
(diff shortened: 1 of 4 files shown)


jonathunne (approved):
Looks good.

PR #63, opened 2026-08-11
refactor(multisig): make message-signing an optional capability interface

AlonzoRicardo:
## Summary

Make **multisig message-signing an optional capability**, instead of requiring it of every implementer. Message-signing (`proposeMessage` / `approveMessageProposal` / `getMessageProposal` / `getMessageProposals`) moves out of the core `IWalletAccountReadOnlyMultisig` / `IWalletAccountMultisig` into two optional interfaces, mirroring `IMultisigOwnerManagement`.

## Motivation

The third multisig implementation — **Solana Squads** (`wdk-protocol-multisig-squads`) — surfaced that message-signing isn't universal. A multisig account is a **keyless** program-controlled address (a PDA on Solana, a contract on EVM), so it can't sign as itself. EVM Safe works around this with **EIP-1271** (a standard letting a keyless contract validate a "the multisig approved this message" signature). **Solana has no EIP-1271 equivalent** and Squads has no message-approval primitive — so the Squads module can only stub-throw `NotSupported` on every message method. Forcing methods onto an implementer that can never fulfil them is the wrong contract; message-signing should be opt-in.

## Changes

- New `src/multisig/multisig-message-signing.js`:
  - `IMultisigReadOnlyMessageSigning` — `getMessageProposal`, `getMessageProposals` (optional, read-only).
  - `IMultisigMessageSigning extends IMultisigReadOnlyMessageSigning` — adds `proposeMessage`, `approveMessageProposal` (optional, writable).
  - Owns the `MultisigMessageProposal` and `MultisigSignature` typedefs.
- Removed those four methods + two typedefs from the core `IWalletAccountReadOnlyMultisig` / `IWalletAccountMultisig`.
- Updated the `/multisig` barrel and the hand-maintained `.d.ts`.

## Impact on implementers

- **EVM Safe:** read-only class `@implements IMultisigReadOnlyMessageSigning`, writable class `@implements IMultisigMessageSigning` — behaviour unchanged.
- **Solana Squads:** implements neither → the message methods (and their `NotSupported` throws) simply go away.
- Breaking to the interface surface, but pre-GA beta; no released implementation depends on these being on the *core* interfaces.

## Out of scope: `verify`

Squads also can't implement `verify(message, signature)` (same reason — no EIP-1271, no signature attributable to a keyless multisig). But `verify` lives on the **base `IWalletAccountReadOnlySimple`**, shared by *every* wallet account — not just multisigs. Making it optional touches the whole wallet-account hierarchy, a much larger blast radius. To keep this PR atomic to multisig message-signing, `verify` is left as-is and tracked separately; Squads keeps throwing on `verify` until that's addressed.

## Testing

`npm run lint` clean, `npm test` 19/19. Interfaces are abstract (`NotImplementedError` stubs), so there are no direct interface tests; the change is verified structurally (exports + inheritance).

### DIFF
--- a/src/multisig/multisig-message-signing.js
+++ b/src/multisig/multisig-message-signing.js
@@ -1,0 +1,90 @@
+// Copyright 2024 Tether Operations Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+'use strict'
+
+import { NotImplementedError } from '../errors.js'
+
+/** @typedef {import('../errors.js').NoSuchElementError} NoSuchElementError */
+
+/**
+ * @typedef {Object} MultisigMessageProposal
+ * @property {string} messageId - The message's hash.
+ * @property {string} message - The original message.
+ * @property {number} confirmations - The current number of confirmations.
+ * @property {number} threshold - The minimum amount of confirmations to sign the message.
+ * @property {string | null} combinedSignature - The final combined signature when the threshold is met.
+ */
+
+/**
+ * @typedef {Object} MultisigSignature
+ * @property {string} signature - The caller's signature.
+ */
+
+/**
+ * Adds the read-only message-signing queries to a multisig wallet.
+ *
+ * @interface
+ */
+export class IMultisigReadOnlyMessageSigning {
+  /**
+   * Returns a list of message proposals by their hashes.
+   *
+   * @param {string[]} messageIds - The list of message hashes
+   * @returns {Promise<Record<string, MultisigMessageProposal | null>>} For each message hash, the message details or
+   *   null if the message has not been found.
+   */
+  async getMessageProposals (messageIds) {
+    throw new NotImplementedError('getMessageProposals(messageIds)')
+  }
+
+  /**
+   * Returns a message proposal by its identifier.
+   *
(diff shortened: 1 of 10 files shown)


src/multisig/multisig-message-signing.js
Davi0kProgramsThings, line 1: Split this file into:
1. src/multisig/multisig-message-signing-read-only.js
2. src/multisig/multisig-message-signing.js

src/multisig/multisig-message-signing.js
```diff
@@ -0,0 +1,90 @@
...
+ * @typedef {Object} MultisigSignature
+ * @property {string} signature - The caller's signature.
+ */
+
+/**
+ * Adds the read-only message-signing queries to a multisig wallet.
+ *
+ * @interface
+ */
+export class IMultisigReadOnlyMessageSigning {
```
Davi0kProgramsThings: Rename this class to:
```suggestion
export class IMultisigMessageSigningReadOnly {
```

Davi0kProgramsThings (changes requested):

types/src/multisig/multisig-message-signing.d.ts
```diff
@@ -0,0 +1,75 @@
...
+     */
+    combinedSignature: string | null;
+};
+export type MultisigSignature = {
+    /**
+     * - The caller's signature.
+     */
+    signature: string;
+};
+export type NoSuchElementError = import("../errors.js").NoSuchElementError;
```
sontuphan: because the `NoSuchElementError` is published in the `index.js`, it seems redundant here to re-export `NoSuchElementError`.

sontuphan (dismissed):

types/src/multisig/multisig-message-signing.d.ts
Davi0kProgramsThings: That's how tsc generates type definitions starting from jsdoc. This is not really a bug though. It makes sense that each files export all of its public api which in this case also includes this error. We are doing this basically everywhere in the type definitions so i wouldn't really change it here.

Davi0kProgramsThings (approved):
LGTM.

PR #66, opened 2026-08-24
refactor(multisig)!: rename MultisigAutoExecuteResult to MultisigInteractionResult

AlonzoRicardo:
Stacked on #65 (retarget to `main` once #65 merges).

## Why

After #65 restores the `transaction` result, the type is still named and documented as auto-execute-specific. It isn't — it carries **the on-chain transaction a write call produces, whenever that call itself settles on-chain**. Per @sontuphan's proposal on the parallel multisig module, one shared field with one meaning serves every backend instead of a bespoke per-chain result:

- Off-chain approval collection, settle on execution → `transaction` is the execution (absent until then).
- Every proposal/approval is its own on-chain transaction → `transaction` is that call's receipt.

## Change

- Rename exported type `MultisigAutoExecuteResult` → `MultisigInteractionResult` (all 8 sites: `propose`/`approveProposal` returns, `index.js` + both `.d.ts` mirrors). Shape unchanged (`{ transaction?: TransactionResult }`) — no type change, per the proposal. `rejectProposal` stays plain.
- Generalize the JSDoc and **pin the contract that defuses the footgun**: the presence of `transaction` does **not** imply execution; execution is determined solely by `status`. (Without this, an EVM consumer would misread a `pending` result that carries a `transaction` as executed.)

## Breaking change

Renames a published exported type. Justified: the old name mischaracterizes the generalized field, and the type is currently orphaned on `main` (post-beta.17 regression), so real-world breakage is minimal — #65 and this land together.

## Verification

- `standard` lint clean
- `jest` 27/27 pass

cc @quocle108 @Davi0kProgramsThings @jonathunne — this is the interface-level home for the widen + rename discussed on the Squads thread; settling it here (like message-signing in #63) rather than on that PR.


sontuphan (dismissed):
LGTM

sontuphan (approved):

